PR #137, opened 2023-11-03
CICD git workflow for prod deploy

prxt6529:


.github/workflows/build-upload-deploy-prod.yml
```diff
@@ -0,0 +1,116 @@
...
+          sleep 30 # Initial sleep for the new EC2 to initialize
+          while : ; do
+            ENV_HEALTH=$(aws elasticbeanstalk describe-environments --environment-names ${{ env.BEANSTALK_ENV_NAME }} --query 'Environments[0].Health' --output text)
+            ENV_STATUS=$(aws elasticbeanstalk describe-environments --environment-names ${{ env.BEANSTALK_ENV_NAME }} --query 'Environments[0].Status' --output text)
+            echo "Environment health: $ENV_HEALTH"
+            echo "Environment status: $ENV_STATUS"
+            if [[ $ENV_HEALTH == "Green" && $ENV_STATUS == "Ready" ]]; then
+              echo "Environment is healthy and ready!"
+              break
+            elif [[ $COUNTER -eq $MAX_RETRIES ]]; then
```
TarmoKalling, line 88: what if during those 30 tries aws does: deploy -> pending -> failed -> rollback -> success
wouldn't we think then that all's good but it really isn't?

prxt6529: i believe you would see the logs in git workflow but any suggestions on this?

TarmoKalling: As a countermeasure we could ask the currently deployed version from eb and compare (as the last step)

TarmoKalling (approved):

PR #145, opened 2023-11-13
Profile ssr init

analyticsflowee:


components/user/UserPage.tsx
```diff
@@ -1,26 +1,61 @@
...
+          balance: gradient.balance,
+          contract: GRADIENT_CONTRACT,
+        });
+      }
+    }
+
+    return cards;
+  };
+
+  const consolidatedOwned = getConsolidatedOwned();
```
prxt6529, line 57: if you buy a card now
it will not be in this `consolidatedTDH.memes` or `consolidatedTDH.gradients` array
those contain data as per last tdh calculation

analyticsflowee: Changed to old approach

prxt6529 (approved):

PR #281, opened 2024-02-09
Not sending x-6529-origin-path anymore

GelatoGenesis:


services/api/common-api.ts
```diff
@@ -1,6 +1,7 @@
 import Cookies from "js-cookie";
 import { API_AUTH_COOKIE } from "../../constants";
 import { getAuthJwt } from "../auth/auth.utils";
+import {headers} from "next/headers";
```
prxt6529, line 4: is this being used

TarmoKalling: removed it

prxt6529 (approved):

PR #544, opened 2024-10-03
Adds playwright tests for most unauthenticated pages.

brookr:


prxt6529:
gmgm @brookr 

some comments

- runs nicely on localhost too which is good to see
- this test is not passing for me "Home Page › should display the Latest Drop section" i dont get why tho its definitely there and it fails both on staging and local

also, its mostly checking titles and pages etc right? no actual functionality - some suggestions (not sure if this is possible)
- when no password -> navigated to /access
- when restricted country -> navigated to /restricted

playwright.config.ts
```diff
@@ -0,0 +1,27 @@
+import { PlaywrightTestConfig, devices } from '@playwright/test';
+import * as dotenv from 'dotenv';
+
+dotenv.config(); // Loads variables from .env
+dotenv.config({ path: '.env.local' }); // Overrides or adds variables from .env.local
```
prxt6529, line 5: can this not be hardcoded to '.env.local'

i remember something about nextjs not supporting .local and my local is actually called '.env.development'

can it be wildcard or support multiple .env ?

brookr: oh yes, this should actually be loading `env.test`, so we can all keep it distinct from dev. i'll update it.

brookr:
> * this test is not passing for me "Home Page › should display the Latest Drop section" i dont get why tho its definitely there and it fails both on staging and local

what does the failing test report? 

> 
> also, its mostly checking titles and pages etc right? 

right, this first set is "without connecting, pages exist and have the expected content". 

> * when no password -> navigated to /access
> * when restricted country -> navigated to /restricted

good thoughts, we can look at adding that in.

prxt6529:
imHo i would like to see the two redirect tests - at least the one with no password taking you to /access should be easy to put in place

also - what is the idea for running this? i mean will they need to be ran manually or will it be part of the build process?

in a perfect world we would want to deploy the changes in an isolated env, run the tests against that and then deploy to prod if successful

brookr:
> imHo i would like to see the two redirect tests - at least the one with no password taking you to /access should be easy to put in place

added to the '/access' page: now testing if logged-in /access shows a disabled input, and logged-out / redirects to /access. 

> also - what is the idea for running this? i mean will they need to be ran manually or will it be part of the build process?

They should be part of the deployment process, yes. Hopefully we all run manually (individual files) and add tests while doing dev work. 

> in a perfect world we would want to deploy the changes in an isolated env, run the tests against that and then deploy to prod if successful

right, prod deploy should be gated on tests passing against staging.

brookr:
SonarCloud issues resolved!

prxt6529:
> right, prod deploy should be gated on tests passing against staging. We can keep working towards easier deployment of the full application stack, so tests can have their own env.

i might be working on branchA, simo is on branchB
both branches are in staging

when i want to merge branchA to prod, it doesn't necessarily mean tests will be successful on staging
and any failures don't mean they are coming from the branch im about to merge

basically staging contains more things than what i want to merge so running tests against staging isn't ideal i think

brookr:
Let’s sort out the deployment flow in another thread. Would love your comments on what I suggested in discord. 

anything else this pr needs for approval?

prxt6529 (approved):

PR #800, opened 2025-05-05
Fix react-tweet video playback

kwizzn:
This allowlists https://video.twimg.com as a media source, enabling video playback in react-tweet `<Tweet id={result.tweetId} />` containers.

Unless there was a good reason to restrict it?


kwizzn:
@ragnep, @prxt6529, @punk6529 any thoughts?

prxt6529:
gmgm @kwizzn thank you for the PR, its a good addition
I will actually add this in another PR 
https://github.com/6529-Collections/6529seize-frontend/pull/806/files#diff-882b2c04b01b2e8b2cdcf1817c30ea503a8005f1c0d54cfff37053b6801fea85

where I am adding `https://*.twimg.com;` to cover any other media sources in the twimg.com domain!

kwizzn:
Nice, and we're live. Thx!

PR #854, opened 2025-05-18
Implement basic waves list virtualization

punk6529:
## Summary
- add `ScrollPositionContext` for storing scroll positions
- create `useVirtualizedWaves` hook to render a limited window of wave rows
- apply virtualization in `UnifiedWavesListWaves`
- test hook logic with 1000 dummy items
- wire IntersectionObserver to sentinel with forwarded refs

## Testing
- `npm test` *(fails: jest not found)*


punk6529:
Needs some fixes as noted below

<html>
<body>
<html><head></head><body><h3>Quick Ship-Check on the latest diff</h3>

Status | Area | Notes
-- | -- | --
✅ | Sentinel attachment | listRef.current?.sentinelRef.current in the deps makes the IntersectionObserver re-register as soon as the sentinel appears. Infinite scroll now works in manual testing.
✅ | Ref hand-off | forwardRef + useImperativeHandle removes the ref-mutation hack—parent always receives live container & sentinel refs.
⚠️ | Effect deps lint | ESLint’s react-hooks/exhaustive-deps will complain about listRef.current?.sentinelRef.current because it’s a property access inside the array. Either disable the line or wrap refs in state (`const [sentinel, setSentinel] = useState<HTMLElement
⚠️ | Loader gating | Loader now shows only when isFetchingNextPage is true. That’s fine for page > 1, but the very first page fetch (initial mount) will no longer display any loading indicator. Decide if UX is acceptable; if not, pass hasNextPage back in and keep the old condition.
🟡 | Initial blank flash & resize | viewportHeight still reads clientHeight only once, so the <100 ms white flash on first mount and the window-resize gap are still there. Not a blocker for functionality, but users will notice on fast navigations / orientation changes. Leave the ticket open.
🟡 | Test looseness | Unit test still allows ≤ 120 rows regardless of viewport. Won’t fail if virtualization silently breaks. OK for now, but tighten later.
🟢 | Risk of regressions | No blocking regressions found in manual scroll / pin / hover testing.


<hr>
<h4>Can you ship?</h4>
<p><em>Yes, if you’re comfortable shipping with the minor UX issues above.</em></p>
<ul>
<li>
<p>Functionally critical bugs (sentinel missing, ref staleness) are fixed.</p>
</li>
<li>
<p>Remaining items are visual polish or edge-case UX; they’re tracked in tickets.</p>
</li>
</ul>
<p><strong>Recommended micro-patch before merge</strong></p>
<pre><code class="language-ts">// UnifiedWavesList.tsx  – silence the lint rule cleanly
const sentinel = listRef.current?.sentinelRef.current;

useEffect(() =&gt; {
  if (!sentinel || !hasNextPage || isFetchingNextPage) return;

  const obs = new IntersectionObserver(cb, { root: listRef.current?.containerRef.current, rootMargin: '100px' });
  obs.observe(sentinel);
  return () =&gt; obs.disconnect();
}, [sentinel, hasNextPage, isFetchingNextPage]);   // &lt;- simple dep
</code></pre>
<p>Add it if you have five minutes; otherwise note the lint disable and merge.</p>
<hr>
<p><strong>TL;DR</strong><br>
Ship now to unblock the team; keep the “initial blank flash &amp; resize” ticket in the backlog and circle back when you have capacity.</p></body></html>
</body>
</html>

ragnep (approved):

PR #1225, opened 2025-06-11
Fixed image modal overlapping status bar on mobile

ragnep:


components/drops/view/item/content/media/DropListItemContentMediaImage.tsx
```diff
@@ -12,6 +16,8 @@ import {
   ImageScale,
 } from "../../../../../../helpers/image.helpers";
 import { useInView } from "../../../../../../hooks/useInView";
+import Tippy from "@tippyjs/react";
+import "tippy.js/dist/tippy.css";
```
prxt6529, line 20: why is this needed here?!
shoudln't the import in _app.tsx be enough? 
i see this import in other files too i think its redundant!

prxt6529: also... i think simo was saying this this is deprecated and i started using react-tooltip
you can find an example en WaveDropReactions if you want to swap to that

prxt6529 (approved):

PR #1312, opened 2025-07-30
Web3modal wagmi removal

simo6529:


tsconfig.json
```diff
@@ -32,5 +32,5 @@
     "pages/**/*.jsx",
     ".next/types/**/*.ts"
   ],
-  "exclude": ["node_modules", "tests", "e2e"]
+  "exclude": ["node_modules", "tests", "e2e", "**/__tests__/**"]
```
prxt6529, line 35: when you add this, the path alias doesn't work in tests :( and it makes it messy
i've noticed chatgpt adding it consistently not sure why it's needed tho

PR #1345, opened 2025-08-25
Migrate/web3modal to reown appkit

simo6529:


components/providers/WagmiSetup.tsx
```diff
@@ -1,104 +1,168 @@
...
-      connector: Connector,
-      existingConnectors: Connector[]
-    ): boolean => {
-      return !existingConnectors.some(
-        (existing) => existing.id === connector.id
-      );
+  // Create adapter with essential configuration only
+  const createAdapterWithWallets = useCallback((wallets: AppWallet[]): WagmiAdapter => {
+    // Basic validation - let util handle detailed validation
+    if (!CW_PROJECT_ID || !VALIDATED_BASE_ENDPOINT || !adapterManager) {
```
prxt6529, line 63: CW_PROJECT_ID check is a bit redundant, its a constant

components/providers/AppKitAdapterManager.ts
```diff
@@ -0,0 +1,273 @@
...
+            errorType: error.name,
+            message: error.message
+          });
+        }
+        // Re-throw to prevent silent failures
+        throw error;
+      }
+    })
+
+    if (!CW_PROJECT_ID) {
```
prxt6529, line 67: CW_PROJECT_ID check is a bit redundant, its a constant

components/providers/WagmiSetup.tsx
```diff
@@ -1,104 +1,168 @@
...
-import { initWeb3Modal } from "./web3ModalSetup";
+import { WagmiAdapter } from '@reown/appkit-adapter-wagmi'
+import { AppKitAdapterManager } from './AppKitAdapterManager';
+import { CW_PROJECT_ID, VALIDATED_BASE_ENDPOINT } from "@/constants";
+import { AppKitValidationError } from '@/src/errors/appkit-initialization';
+import { Capacitor } from '@capacitor/core';
+import { useAuth } from '../auth/Auth';
+import { sanitizeErrorForUser, logErrorSecurely } from '@/utils/error-sanitizer';
+import {
+  initializeAppKit as initializeAppKitUtil,
```
prxt6529, line 22: this naming and renaming and reusing 'initializeAppKit' as callback is a bit weird and confusing :( 

like i search for `initializeAppKit` in the repo and its hard to actually see where its used and finds other results

components/providers/Providers.tsx
```diff
@@ -30,14 +30,14 @@ export default function Providers({
...
     <QueryClientSetup>
-      <WagmiSetup>
-        <CapacitorSetup />
-        <IpfsImageSetup />
-        <ReactQueryWrapper>
-          <SeizeSettingsProvider>
-            <EmojiProvider>
-              <IpfsProvider>
-                <AppWalletsProvider>
+      <AppWalletsProvider>
```
prxt6529, line 33: it feels kinda strange for AppWalletsProvider to be outside wagmi instead of inside 
it seems to work so i guess its fine just pointing it out

simo6529: yes, it was better and more straight forward to set it up that way, so if we start setting up wagmi, we have easier access to appwalletsprovider.

prxt6529 (approved):

PR #1428, opened 2025-09-16
feat: improve filter target accessibility

punk6529:
## Summary
- replace the button-wrapped radios with a semantic fieldset, legend and individual radio inputs to enable native keyboard support
- ensure each radio shares a unique group name generated per render so only the intended option can be selected at once
- update the CommonFilterTargetSelect tests to cover the new group label and keyboard navigation behaviour

## Testing
- npm run lint
- npm run type-check *(fails: repository contains pre-existing type errors in test fixtures)*
- npm run test *(fails: many suites require BASE_ENDPOINT env var that is not configured in this environment)*

------
https://chatgpt.com/codex/tasks/task_e_68c9430aa9a08321ab6907bf81932241


prxt6529:
WIll be done in https://github.com/6529-Collections/6529seize-frontend/pull/1497

PR #2362, opened 2026-05-08
wip

simo6529:
## Summary by CodeRabbit

* **Tests**
  * Added tests covering “no active proxy” scenarios: verifies read-marking uses the connected profile’s auth when roles match, and that it skips/send-none when roles disagree (no API call when queued-off).

* **Refactor**
  * Authentication selection for marking wave notifications as read now accounts for connected profiles and normalizes role handling to determine whether to use connected-profile auth or a proxy-derived identity.


simo6529:
@coderabbitai full review

ragnep (approved):

PR #2553, opened 2026-06-09
Harden CSP source policy

punk6529:
## Summary

This reviews the broad CSP allowances in `config/securityHeaders.ts` and tightens the parts that were not justified:

- Replace `connect-src *` with scheme-limited `https:` and `wss:` plus the existing explicit sources. The app still needs broad HTTPS/WSS connectivity for wallet/RPC providers, presigned uploads, analytics, and user/decentralized media probes, but a raw wildcard also permits broader schemes than necessary.
- Replace `img-src *` with `https:` while preserving `data:`, `blob:`, and `ipfs:`. The app displays user-supplied NFT/drop/profile media from decentralized or arbitrary hosts, but production HTTP image loading is not justified.
- Change `object-src data:` to `object-src 'none'`. I found PDF/interactive rendering uses iframes and 3D rendering uses `model-viewer`; I did not find an app dependency on `<object>` or `<embed>`.
- Move `unsafe-eval` behind an explicit local/dev option. Source and selected dependency scans did not show a production need for eval-like execution; local/dev keeps the compatibility escape hatch.
- Keep `unsafe-inline` with inline comments documenting the current need: Next/bootstrap behavior, consent-gated GA injection, and imported legacy WordPress-style pages with inline CSS/script surfaces.

## Verification

- `6529 run test -- __tests__/config/securityHeaders.test.ts`
- `6529 run lint:diff`
- `6529 run lint:changed`

## Summary by CodeRabbit

* **Security Improvements**
  * Strengthened Content-Security-Policy: removed permissive wildcards, require secure origins for APIs/gateways, disallow plain-HTTP CDN, and only allow IPFS gateway when HTTPS; localhost exceptions require explicit enablement.

* **Bug Fixes**
  * object-src corrected to block object embedding.

* **Configuration**
  * 'unsafe-eval' only enabled when explicitly allowed (e.g., dev/local).

* **Tests**
  * Added CSP tests for directives, gateway handling, localhost allowances, and unsafe-eval behavior.


punk6529:
Reviewed the Sonar hotspot and agreed it was justified. The clear-text cdnjs style source was only supporting protocol-relative legacy imports on a local HTTP dev server, but production CSP should not allow HTTP CDN styles for that convenience.

Updated the PR to remove that source from `style-src` and added a regression test asserting CSP style sources do not allow `http:` entries. Re-ran the focused CSP test, `lint:changed`, and `typecheck:changed` successfully.

ragnep (approved):

PR #2678, opened 2026-06-16
[codex] Add delegation content package and IPFS loader

punk6529:
## Issue

Delegation help article bodies were effectively remote mutable content. We need reviewed repo content as the source of truth, with IPFS as canonical publish storage and S3/CloudFront only as an acceleration mirror.

## Fix

Adds a reviewed delegation docs package, manifest-based delivery, verified runtime loading, and source-of-truth docs for the IPFS -> CDN delivery model.

## Changes

- Adds checked-in versioned delegation article HTML and packaged article assets.
- Keeps article metadata in `content/delegation/manifest.json` and source assets in `content/delegation/assets`.
- Uses one reviewed article-body copy in `public/delegation-content/{version}/html` so the runtime fallback is fetchable without duplicating HTML bodies in the repo.
- Generates a public versioned content bundle plus manifest with article and asset hashes.
- Adds a runtime loader that prefers version/CID-addressed CDN/IPFS URLs and verifies article SHA-256 before rendering.
- Caps article response bodies before rendering, including `content-length` and stream-reader guards for untrusted gateways.
- Rewrites all relative article `src`/`href` references against the verified article URL and fails closed if called outside a DOM document.
- Keeps mutable legacy S3 import behind explicit `DELEGATION_DOCS_IMPORT_LEGACY_S3=1` seed-import mode.
- Rewrites packaged article asset refs to local package paths and app article links to app routes.
- Derives the builder article catalog from the reviewed manifest instead of a duplicate static JS table.
- Adds page metadata helpers and content renderer tests.
- Adds source-of-truth and content-delivery docs under `ops/docs/delegation/`.
- Fixes Sonar's regex hotspot warnings in the URL/heading normalizers and removes the duplicate article-body package shape that was driving the duplication gate.

## Validation

- `node ops\scripts\build-delegation-docs-content.mjs`
- Blank-target link rel invariant check passed
- Packaged HTML S3 URL scan passed
- Versioned package asset count: 54 files, mirrored from 54 reviewed source asset files
- `seize run test:no-coverage -- __tests__/app/delegation.section.test.ts __tests__/components/delegation/html/delegationContent.test.ts __tests__/components/delegation/html/DelegationHTML.test.tsx`
- `seize run lint:changed`
- `codex-diff-check`

## Risk

Medium. This introduces a new content delivery path, but runtime hash verification, response-size guards, and local packaged fallback keep failure modes bounded.

## Review Notes

Stacked on #2677. Review this PR against `codex/fix-bootstrap-sass-build` to avoid the Sass build-fix diff. The package now keeps reviewed repo content as source, IPFS as canonical publish storage, and S3/CloudFront as a CID/version-addressed acceleration layer.


punk6529:
@coderabbitai review

PR #2704, opened 2026-06-17
[codex] Fix wave score tooltip layering

punk6529:
## Summary
- Raise the rich Wave Score tooltip above the wave stream so it is actually visible on hover.
- Replace the calculator placeholder with a live exact wave-name example (`Memes-Chat`) instead of the non-matching lowercase sample.

## Why
Staging E2E showed the active-wave score tooltip had the right content and dark styling in the DOM, but it was painted underneath the wave content. The calculator also worked with exact wave names/URLs/ids, but the placeholder example did not match staging data.

## Validation
- `seize run test:no-coverage -- __tests__/components/waves/WaveTrustSignals.test.tsx __tests__/app/networkPages.test.tsx --runInBand`
- `seize run lint:single -- components/waves/WaveTrustSignals.tsx components/waves/discovery/WaveScoreTransparencyPage.tsx`
- `codex-diff-check -- components/waves/WaveTrustSignals.tsx components/waves/discovery/WaveScoreTransparencyPage.tsx`
- Authenticated staging Playwright debug confirmed the z-index adjustment makes the tooltip visible above the wave stream.

### DIFF
--- a/components/waves/WaveTrustSignals.tsx
+++ b/components/waves/WaveTrustSignals.tsx
@@ -465,7 +465,7 @@ function WaveScoreSummaryTooltip({
     <span
       id={id}
       role="tooltip"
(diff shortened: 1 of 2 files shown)


punk6529:
Addressed the 6529bot general finding in the current head `b160e6341`: the score chip keeps `tw-isolate`, and the elevated `tw-z-[10000]` is scoped to the tooltip element itself. I also verified `Memes-Chat` resolves on staging during authenticated calculator E2E. Current checks are green.

PR #2747, opened 2026-06-18
Move subwave chevron beside wave names

punk6529:
## Summary
- Move the subwave expand chevron inline to the right of the wave name so parent row avatars keep their vertical alignment.
- Restyle the control with a stronger icon stroke, clearer active state, and focus-visible treatment.
- Update app and web sidebar row tests for the new link/control structure.

## Testing
- `seize run format:changed`
- `seize run lint:changed`
- `seize run test:no-coverage -- __tests__/components/brain/left-sidebar/waves/BrainLeftSidebarWave.test.tsx __tests__/components/brain/left-sidebar/web/ExpandedWave.test.tsx --runInBand`
- `git diff --check --ignore-cr-at-eol main...HEAD`
- Local Playwright visual QA against mocked sidebar API data: parent wave title x positions remained aligned, chevron sits directly after the title, and expanded child rows retain their nested indentation.

## Local note
- `seize run quality:changed` / `seize run check:changed` currently fail in this Windows Codex worktree before the inner formatter prints output because `scripts/quality.js` invokes the repo-local `bin\6529.cmd`, whose Bash path handoff cannot open `D:\...\bin\6529` here. The same formatter, lint, typecheck, tests, and whitespace gates pass through the supported local `seize` wrapper.

### DIFF
--- a/__tests__/components/brain/left-sidebar/waves/BrainLeftSidebarWave.test.tsx
+++ b/__tests__/components/brain/left-sidebar/waves/BrainLeftSidebarWave.test.tsx
@@ -263,15 +263,10 @@ describe("BrainLeftSidebarWave", () => {
     const titleLink = screen.getByRole("link", { name: "Chat Wave" });
     expect(titleLink.nextElementSibling).toBe(expandButton);
     expect(expandButton.parentElement).toContainElement(titleLink);
-    expect(
-      screen.getByRole("link", { hidden: true, name: "" })
-    ).toHaveAttribute("aria-hidden", "true");
-    expect(
-      screen.getByRole("link", { hidden: true, name: "" })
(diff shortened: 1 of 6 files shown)


punk6529:
Addressed the 6529bot general accessibility findings in `299a61bf091401909321233373f964de68c838a5`:

- The avatar is now a presentational `<div aria-hidden="true">`, not a hidden clickable link.
- The title is the only real wave navigation link, so there is no duplicate interactive path.
- Tests now use `data-testid="sidebar-wave-avatar"` and assert the avatar is not inside an anchor.

The async `canExpand` note is acceptable for this UI: the title/avatar alignment stays fixed, and the chevron appears inline after the title rather than reserving space before the avatar.

PR #2755, opened 2026-06-18
[codex] Guard prod deploy workflow ref

punk6529:
## Summary
- fail the production web deploy workflow before build/upload/deploy when the selected ref is not `main`
- add a `web-deploy-prod` concurrency group so production deploys queue instead of overlapping
- attach the deploy job to the GitHub `production` environment, now configured with a `main` branch deployment policy
- limit workflow `GITHUB_TOKEN` permissions per CodeQL feedback

## Why
A manual `Web Deploy - PROD` run was dispatched from `codex/prod-subwave-chevron-2747`. The SHA was already merged into `main`, but the workflow should not permit production deploys from any non-main selected ref.

## Validation
- `codex-diff-check -- .github/workflows/build-upload-deploy-prod.yml`
- parsed `.github/workflows/build-upload-deploy-prod.yml` with PyYAML
- confirmed guard text, production environment, concurrency group, and explicit permissions are present

### DIFF
--- a/.github/workflows/build-upload-deploy-prod.yml
+++ b/.github/workflows/build-upload-deploy-prod.yml
@@ -2,6 +2,10 @@ name: Web Deploy - PROD
 
 on: workflow_dispatch
 
+concurrency:
+  group: web-deploy-prod
+  cancel-in-progress: false
+
 env:
   AWS_BUCKET: "seize-web"
   COMMIT_SHA: "${{ github.sha }}"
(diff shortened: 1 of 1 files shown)


punk6529:
Confirmed the final workflow file includes both permissions blocks: permissions: {} on ssert-main-ref and permissions: contents: read on uild-upload-deploy. Re-ran codex-diff-check and PyYAML parsing after the rebase.

PR #2760, opened 2026-06-18
[codex] Fix CMS validator asset reference lint

punk6529:
## Summary
- Fix the production build blocker from `sonarjs/no-identical-functions` in the CMS V1 validator
- Keep required asset-reference behavior unchanged by validating missing assets directly instead of routing through the optional-reference helper

## Why
The latest PROD deploy failed during `./bin/6529 run build` because full `eslint . --quiet` flagged `lib/profile-cms/protocol/v1/validation.ts`. This is unrelated to the wave chevron/card polish, but it blocks deploying current `main`.

## Validation
- `seize run lint:quiet`
- `seize run test:no-coverage -- __tests__/lib/profile-cms/protocol/v1/fixtures.test.ts __tests__/lib/profile-cms/protocol/v1/hash.test.ts`
- `codex-diff-check`

### DIFF
--- a/lib/profile-cms/protocol/v1/validation.ts
+++ b/lib/profile-cms/protocol/v1/validation.ts
@@ -640,7 +640,7 @@ function validateNftMediaProfiles(
     }
 
     profile.original_asset_ids?.forEach((assetId, assetIndex) => {
-      validateAssetReference(
+      validateOptionalAssetReference(
         assetId,
         assetMap,
         `${path}/original_asset_ids/${assetIndex}`,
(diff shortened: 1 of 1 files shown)


punk6529:
Updated per 6529bot review:

- removed `validateAssetReference` entirely instead of changing its semantics
- routed all former call sites through the existing `validateOptionalAssetReference` helper
- this preserves the previous falsy/empty-string behavior and removes the duplicate helper implementation that triggered `sonarjs/no-identical-functions`

Re-ran:
- `seize run lint:quiet`
- `seize run test:no-coverage -- __tests__/lib/profile-cms/protocol/v1/fixtures.test.ts __tests__/lib/profile-cms/protocol/v1/hash.test.ts`
- `codex-diff-check`

PR #2761, opened 2026-06-18
Fix frontend production build loaders

punk6529:
## Summary
- Resolve Bootstrap SCSS through its explicit package path so Turbopack/Sass does not match unrelated npm package files during builds.
- Add a CommonJS `next-sitemap` build shim so the deploy `postbuild` step can load the existing typed sitemap config.

## Validation
- `seize run lint:changed`
- `seize run test:no-coverage -- --testMatch "**/__tests__/sitemap/nextSitemapConfig.test.ts"`
- `codex-diff-check`
- `seize run build`

Part of release train `profile-cms-phase-0-8`.


punk6529:
/6529bot review

Please focus on build/deploy safety: Bootstrap Sass resolution under Turbopack, the `next-sitemap` CommonJS shim for the existing typed config, and whether this could affect runtime behavior.

punk6529:
Updated the hotfix after review:

- Moved `tsx` from `devDependencies` to `dependencies` so the postbuild sitemap shim still works after production pruning.
- Added a guard in `next-sitemap.build.cjs` so the build fails loudly if the typed sitemap config does not export an object.
- Tightened `pnpm-lock.yaml` to the minimal importer move for `tsx` instead of keeping install-time dedupe churn.

On the Bootstrap import: I tried replacing the explicit path with package/importer-based resolution (`sassOptions.loadPaths`, Sass `NodePackageImporter`/`pkgImporter`, and a custom importer). Under the current Next 16/Turbopack/Sass loader path, those either still resolve Bootstrap's legacy internal `@import "progress"` to the npm `progress` package or are rejected by the loader. The explicit `../node_modules/bootstrap/scss/bootstrap` import is the only verified option that keeps Bootstrap partials resolving inside Bootstrap and preserves the production build.

Local verification on commit `b6bd21e91`:
- `seize run lint:changed`
- `seize run test:no-coverage -- --testMatch "**/__tests__/sitemap/nextSitemapConfig.test.ts"`
- `codex-diff-check`
- `seize install:frozen`
- `seize run build`

/6529bot review

PR #2768, opened 2026-06-18
Fix missing profile CMS site metadata fallback

punk6529:
## Summary

Fixes a production smoke failure after the Profile CMS release train: `/punk6529/index.html` rendered the CMS error boundary (`Website unavailable`) when no primary CMS package exists.

Root cause: `generateMetadata` called `notFound()` when the profile had no primary CMS site. In production this can surface as a Server Components render error before the page returns its own safe no-site/404 behavior.

Changes:
- Return generic CMS empty-state metadata when there is no primary CMS package.
- Add a route regression test proving metadata generation does not throw for missing primary CMS sites.

## Validation

- `seize run test:no-coverage -- __tests__/app/profile-cms-route.test.tsx --runInBand` via temporary Jest path workaround: 7 tests passed
- `seize run lint:changed`
- `codex-diff-check`

## Production Context

This is a fix-forward after production deploy `2b27ee1e` completed successfully but smoke found `/punk6529/index.html` rendering `Website unavailable` instead of a safe missing-site result. Builder remains hidden in production as 404 when CMS builder flags are off.

### DIFF
--- a/__tests__/app/profile-cms-route.test.tsx
+++ b/__tests__/app/profile-cms-route.test.tsx
@@ -1,6 +1,8 @@
 import { render, screen } from "@testing-library/react";
 
-import ProfileCmsPage from "@/app/[user]/[...cmsPath]/page";
+import ProfileCmsPage, {
+  generateMetadata,
+} from "@/app/[user]/[...cmsPath]/page";
 import exhibitionRoomPackage from "@/ops/workstreams/profile-native-cms-roadmap/phase-1/fixtures/valid/exhibition-room.package.json";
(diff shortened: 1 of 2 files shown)


punk6529:
/6529bot review

Production fix-forward for Profile CMS release train. Smoke on `https://6529.io/punk6529/index.html` after deploy `2b27ee1e` hit the CMS error boundary when no primary CMS package exists. This patch keeps page-level no-site/404 behavior but makes `generateMetadata` return generic metadata instead of throwing `notFound()`.

Validation:
- Focused route regression test passed: 7/7
- `seize run lint:changed`
- `codex-diff-check`

PR #2769, opened 2026-06-18
Render empty CMS state for missing primary site

punk6529:
## Summary

Follow-up fix-forward for the Profile CMS production smoke failure.

PR #2768 removed the metadata `notFound()` throw, but production smoke still showed `/punk6529/index.html` rendering the CMS error boundary. This patch removes the page-level throw for the valid CMS route/no-primary-site case and renders the explicit CMS empty state instead.

Behavior preserved:
- Non-CMS profile paths like `/punk6529/collected` still return not-found.
- Published packages with missing internal CMS routes still render the CMS empty state.
- Builder remains gated by production flags.

## Validation

- Focused route regression test via temporary Jest config workaround: 8/8 passed
- `seize run lint:changed`
- `codex-diff-check`

## Production Context

Production deploy `6bf98060` still rendered `Website unavailable` at `/punk6529/index.html`. This PR should make missing primary CMS packages render `Website page not found` instead of tripping the route error boundary.

### DIFF
--- a/__tests__/app/profile-cms-route.test.tsx
+++ b/__tests__/app/profile-cms-route.test.tsx
@@ -119,17 +119,19 @@ describe("profile CMS App Router catch-all", () => {
     );
   });
 
-  it("returns a safe not-found when no primary CMS site exists", async () => {
+  it("returns a safe empty state when no primary CMS site exists", async () => {
     getProfileCmsPrimarySiteMock.mockResolvedValueOnce(null);
 
-    await expect(
-      ProfileCmsPage({
-        params: Promise.resolve({
-          user: "punk6529",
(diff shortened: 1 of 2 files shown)


punk6529:
/6529bot review

Follow-up production fix-forward. #2768 fixed the metadata path, but production smoke on deploy `6bf98060` still showed `/punk6529/index.html` rendering the CMS error boundary. This patch renders `ProfileCmsEmptyState` for valid CMS routes when no primary package exists, while preserving `notFound()` for non-CMS profile paths.

Validation:
- Focused route regression test: 8/8 passed
- `seize run lint:changed`
- `codex-diff-check`

PR #2876, opened 2026-06-24
[codex] Center mobile dock active pill

prxt6529:
## Summary
- position the floating dock active pill from the active nav item's measured center instead of a coarse index percentage
- add more horizontal breathing room at the dock edges and trim the active pill width so first/last active icons do not crowd the rounded dock ends
- keep a padding/gap-aware CSS fallback for first render and zero-geometry environments
- cover the offset case in BottomNavigation tests so the pill follows real item centers

## Validation
- ./bin/6529 run test -- __tests__/components/navigation/BottomNavigation.test.tsx __tests__/components/navigation/NavItem.test.tsx
- ./bin/6529 run check:changed

### DIFF
--- a/__tests__/components/navigation/BottomNavigation.test.tsx
+++ b/__tests__/components/navigation/BottomNavigation.test.tsx
@@ -67,8 +67,57 @@ const flushAnimationFrame = async () => {
   });
 };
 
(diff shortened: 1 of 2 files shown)


prxt6529:
@coderabbitai review

PR #2883, opened 2026-06-25
[codex] Fix active wallet session verification

prxt6529:
## Summary

Fix the v2 web-session verification path used by share/auth upgrade checks so it verifies the app's active stored wallet account before falling back to the live wallet-provider address.

## Root Cause

In multi-account flows, the active app account and the wallet provider's currently connected address can diverge. After authenticating Wallet B, switching the app back to Wallet A could still verify Wallet B's v2 session state, causing Wallet A to incorrectly show the upgrade-auth prompt even though it had just completed v2 auth.

## Changes

- Prefer `getWalletAddress()` over the live provider `address` when verifying the active v2 web session.
- Add a regression test covering an active stored account that differs from the live provider address.

## Validation

- `./bin/6529 run test:no-coverage -- __tests__/components/auth/Auth.test.tsx --runInBand`
- `./bin/6529 exec eslint --no-warn-ignored --max-warnings=0 components/auth/Auth.tsx __tests__/components/auth/Auth.test.tsx`
- `./bin/6529 run typecheck:ci`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session verification when the wallet provider’s current address differs from the saved connected account.
  * Verification now uses the active wallet address consistently, helping avoid false mismatches during session checks.
  * Added coverage for this scenario to prevent regressions.

### DIFF
--- a/__tests__/components/auth/Auth.test.tsx
+++ b/__tests__/components/auth/Auth.test.tsx
@@ -5,6 +5,8 @@ import Auth, { AuthContext, useAuth } from "@/components/auth/Auth";
 import { ReactQueryWrapperContext } from "@/components/react-query-wrapper/ReactQueryWrapper";
 import { mockTitleContextModule } from "@/__tests__/utils/titleTestUtils";
 import { commonApiFetch, commonApiPost } from "@/services/api/common-api";
+import type * as AuthUtilsModule from "@/services/auth/auth.utils";
+import type * as SessionV2Module from "@/services/auth/session-v2.utils";
 
 const mockQueryClient = {
   getQueryData: jest.fn(),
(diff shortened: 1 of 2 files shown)


prxt6529:
Fresh audit pass after the latest frontend review-fix commit: no new actionable findings.

I re-read the final diff around `components/auth/Auth.tsx` and `__tests__/components/auth/Auth.test.tsx`. The runtime change is still tightly scoped: `ensureActiveSessionV2WebSessionForActiveWallet` now verifies the stored active wallet address before falling back to the live provider address. That matches the multi-account A/B/A failure mode where the app-active account and provider-selected account can diverge.

The regression test now covers that divergence directly and avoids the untyped context/mock setup that CodeRabbit flagged. I do not see a new auth bypass, stale-session, or multi-account regression risk in the final PR diff.

PR #3033, opened 2026-07-04
Add coverage floor on main pushes

punk6529:
## Issue
- Nothing stops global test coverage from eroding: Jest computes coverage but no gate consumes it, so a merge can silently drop coverage. The repo-health campaign requires a monotonic coverage backstop (see `ops/workstreams/repo-health-2026-07/README.md`).

## Fix
- A `Coverage Floor` workflow runs the full Jest suite with coverage on every push to `main` and fails when any global metric drops more than 0.1 points below a checked-in baseline. Rises above tolerance pass with a warning suggesting the one-command baseline bump, so the floor only moves up.

## Changes
- New `.github/workflows/coverage-floor.yml`: push→main + workflow_dispatch, plus a paths-filtered `pull_request` trigger so PRs that change the gate itself (workflow/script/baseline — including this PR and future baseline bumps) validate pre-merge. Read-only permissions, no secrets, installs via the 6529 wrapper + Socket Firewall exactly like app-pr-ci.
- New `scripts/coverage-floor.cjs`: dependency-free comparator over `coverage/coverage-summary.json` totals (lines/statements/functions/branches); `--update` rewrites the baseline.
- New `scripts/coverage-floor-baseline.json`: seeded from a clean-main full-suite run — lines 74.76 / statements 74.39 / functions 70.31 / branches 60.98.
- `ops/workstreams/repo-health-2026-07/run-log.md`: milestone entry (required-checks ruleset step, full-suite timing, known pre-existing red).

## Validation
- Full suite on clean main: 1957 suites / 11018 tests, ~6 min wall with coverage; comparator verified in pass / drop-fails / rise-warns modes against synthetic summaries and the real run.
- `prettier --check` clean; repo workflow-security validator: zero findings; DCO signoffs present.
- This PR triggers its own `Coverage floor` check run (paths filter matches), which validates the seed baseline on ubuntu before merge; if runner coverage differs from the seed beyond tolerance, I will refresh the baseline in this PR from the CI artifact rather than merge red.
- Known pre-existing issue (out of scope, logged + task spawned): `__tests__/hooks/useDownloader.test.ts` fails to load on main yet Jest exits 0; it does not affect this job's exit code.

## Risk
- Level: Low
- Why: additive workflow + scripts; nothing runtime-facing. Worst case is a red post-merge signal on main pushes (non-blocking for the merge itself) or a false alarm from coverage nondeterminism, both fixed by a one-command baseline refresh.
- Rollback: revert the commit; the workflow disappears with the file.

## Review Notes
- Deliberately gates main pushes rather than every PR: the full suite fits (~6 min) but PR latency stays owned by app-pr-ci's risk-plan lanes; the paths-filtered PR trigger covers the one case where pre-merge validation matters (changes to the gate itself).
- `Coverage floor` must NOT be made a required status check — the paths filter means most PRs never create the check run.

## Summary by CodeRabbit

* **New Features**
  * Added automated coverage checks in CI to help keep test coverage from slipping.

* **Bug Fixes**
  * Fixed a test-loading issue so the full test suite runs more reliably.

* **Tests**
  * Added coverage-check test scenarios for matching, dropping, rising, and invalid coverage data.

### DIFF
--- a/.github/workflows/coverage-floor.yml
+++ b/.github/workflows/coverage-floor.yml
@@ -5,19 +5,29 @@ name: Coverage Floor
 # (scripts/coverage-floor-baseline.json), so coverage can only ratchet up.
 # Refresh the baseline with `node scripts/coverage-floor.cjs --update` after
 # running the suite with coverage. Runs post-merge on purpose: PR latency
-# stays owned by the risk-plan lanes in app-pr-ci.yml.
+# stays owned by the risk-plan lanes in app-pr-ci.yml. PRs that change the
+# coverage gate itself (workflow, script, or baseline) also run it so gate
+# changes and baseline bumps are validated before merge. Not a required
+# status check: the paths filter means most PRs never create this check run.
 
 on:
   push:
     branches:
       - main
+  pull_request:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/coverage-floor.yml"
+      - "scripts/coverage-floor.cjs"
+      - "scripts/coverage-floor-baseline.json"
   workflow_dispatch:
 
 permissions:
   contents: read
 
 concurrency:
-  group: coverage-floor-${{ github.ref }}
+  group: coverage-floor-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/coverage-floor.yml
+++ b/.github/workflows/coverage-floor.yml
@@ -1,0 +1,77 @@
+name: Coverage Floor
+
+# Full Jest suite with coverage on every push to main. Fails when global
+# coverage drops more than the tolerance below the checked-in baseline
+# (scripts/coverage-floor-baseline.json), so coverage can only ratchet up.
+# Refresh the baseline with `node scripts/coverage-floor.cjs --update` after
+# running the suite with coverage. Runs post-merge on purpose: PR latency
+# stays owned by the risk-plan lanes in app-pr-ci.yml.
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: coverage-floor-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  coverage-floor:
+    name: Coverage floor
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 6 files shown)


punk6529:
Response to the 6529bot general reviews (8dcd8e2a3f29) — triaged, fixes pushed in 2779937b8:

**Fixed**
- `tolerance_points` from the checked-in baseline is now authoritative in `runCheck` (validated, 0.1 fallback), and `--update` preserves a customized tolerance instead of resetting it. Covered by tests.
- Unguarded `JSON.parse` on summary/baseline replaced with friendly fail-fast messages; corrupt-summary path covered by a test.
- Comparator now has its own Jest suite (8 tests: match, exact-boundary, drop, rise, tolerance override, tolerance preservation, corrupt summary, missing baseline).
- The silent-suite-failure concern: the underlying observation was a measurement artifact on my side (exit code read through a pipe — it returned `tail`'s status). Jest exit codes are correct: suite-load failures and assertion failures both exit 1, so the "Run full Jest suite" step hard-fails the job on any broken suite; no extra assertion layer needed. What WAS real: `__tests__/hooks/useDownloader.test.ts` genuinely failed to load on main — fixed in this PR (local `capacitor-secure-storage-plugin` mock; its 6 tests now run) because a green main suite is a prerequisite for this gate. Full suite: 1957/1957 suites, 11023 tests, exit 0. The in-flight Coverage floor run on the previous head (8dcd8e2a3) is expected to fail on exactly that pre-fix suite — evidence the gate catches it.
- Baseline refreshed from the post-fix run (the fixed suite adds ~0.05 points per metric): lines 74.81 / statements 74.44 / functions 70.34 / branches 61.03.

**Confirmed as designed (documented in workflow + script comments)**
- Exact-tolerance drops pass: the contract is "more than `tolerance_points` points"; now stated explicitly in the script header.
- Post-merge enforcement: intended. Making this a required PR check would either run the ~6-min full suite on every PR (duplicating app-pr-ci's risk-plan lanes and doubling PR latency) or leave a paths-filtered required check that blocks PRs where no run is created — GitHub treats a missing required context as "expected" forever. The floor is a backstop: a red main run pinpoints the offending merge for revert/fix-forward, and the PR-side paths trigger covers changes to the gate itself. If the team later wants pre-merge coverage gating, the right shape is a diff-coverage check inside app-pr-ci, not this workflow.

**Deferred with rationale**
- Sorting baseline keys on write: key order follows the fixed `TRACKED_METRICS` constant, so output is already deterministic; re-sorting alphabetically would churn the committed file for no stability gain.

PR #2642, opened 2026-06-13
Localize profile stats row

punk6529:
## Issue
- Continue the progressive WCAG 2.2 AA and i18n migration on a low-risk user profile surface.
- The profile header stats row still had hardcoded labels and link/button names that were less descriptive for assistive technology.

## Fix
- Move the touched stats-row labels and action names behind the source-locale message dictionary.
- Add explicit accessible names for the TDH, xTDH, NIC, Rep, and followers controls that include the active handle and displayed values.
- Keep this as a component-level migration using the existing `en-US` fallback path; no route-level locale migration is introduced.

## Changes
- Added a profile stats-row message helper and `user.statsRow.*` source messages.
- Updated `UserStatsRow` to use message-backed visible labels and accessible control names.
- Formatted follower counts through the repo i18n integer helper while preserving existing product formatting for TDH/xTDH/NIC/Rep values.
- Added focused tests for link/button accessible names, singular follower copy, rate labels, and fallback message behavior.
- Updated the durable frontend a11y/i18n workstream memory.

## Validation
- `6529 run test -- --runTestsByPath __tests__/components/user/utils/stats/UserStatsRow.test.tsx __tests__/components/user/user-page-header/stats/UserPageHeaderStats.test.tsx __tests__/i18n/messages.test.ts --silent --verbose=false --coverageReporters=none`: passed, 3 suites / 8 tests.
- Targeted ESLint for touched source files: passed.
- `6529 run typecheck:changed`: passed for 112 changed TypeScript files.
- `6529 run react-doctor:diff`: exited 0 at 99/100; only the known unrelated `contexts/EmojiContext.tsx` fetch-in-effect diagnostic remains.
- `git diff --check`: passed; only line-ending warnings on unrelated dirty local files.
- Next MCP `get_errors` on the live local frontend: no config or session errors.
- Browser smoke on `/punk6529` at desktop and 390px mobile widths verified the updated stat control accessible names and practical layout.

## Risk
- Level: Low
- Why: Component-level text/accessibility change on the profile header stats row, with no data fetching, routing, transaction, or dependency changes.
- Rollback: Revert this PR to restore the previous hardcoded labels and action names.

## Review Notes
- This PR is stacked on #2641 and is review-ready only. Per workstream policy, do not merge without human approval.
- Please focus review on accessible names, message-key shape, and whether the values included in the labels are useful without being noisy.

## Summary by CodeRabbit

* **New Features**
[...]

PR #2604, opened 2026-06-11
Add The Memes card accessibility i18n groundwork

punk6529:
## Issue
- The frontend needs progressive WCAG 2.2 AA and i18n migration page by page.
- The first safe implementation target is The Memes list/card surface, without touching mint or transaction flows.

## Fix
- Adds a lightweight repo i18n runtime for the initial locale set and wires The Memes list/card labels through it while keeping the default route locale as `en-US`.
- Improves touched-surface accessibility with explicit card link names, pressed-state sort controls, localized aria labels, a live loading status, a dedicated empty state, a page-level Suspense boundary for `useSearchParams`, and a visible keyboard focus ring on card links.
- Fixes the Windows `react-doctor:diff` wrapper path so the validation command works in this EC2 Windows + WSL2 development setup.

## Changes
- Added `SUPPORTED_LOCALES`, `DEFAULT_LOCALE`, `SupportedLocale`, case-insensitive locale normalization helpers, and native `Intl` helpers.
- Added message dictionaries for `en-US`, `en-GB`, `fr-FR`, `es-ES`, and `de-DE`, with fallback to `en-US`.
- Migrated The Memes list sorting labels, volume labels, card metric text, card aria labels, loading/empty states, and page title context to the message/formatting helpers.
- Wrapped the `/the-memes` page client component in Suspense per the local Next.js `useSearchParams` docs.
- Updated `react-doctor:diff` to resolve Git for Windows and invoke the `react-doctor.cmd` shim through the shell on Windows.
- Updated The Memes list docs and workstream status.

## Validation
- `git diff --check`
- Focused ESLint on owned files via `pnpm exec eslint --no-warn-ignored --max-warnings=0 ...`
- `SEIZE_6529_COMMAND=1 pnpm run typecheck:changed` (passed for 12 changed TypeScript files)
- `pnpm exec jest --runTestsByPath ...` for The Memes card/dropdown/helper tests and i18n helper tests (6 suites, 12 tests passed)
- `SEIZE_6529_COMMAND=1 pnpm run react-doctor:diff` now runs successfully; it reports a pre-existing unrelated diagnostic in dirty `contexts/EmojiContext.tsx`, outside this PR's staged/owned changes.
- `seize-local-dev status` confirmed frontend `http://localhost:3001`, API `http://localhost:3000`, WS `ws://localhost:3000`
- Next MCP `get_errors` on port 3001 after browser verification: no config/session errors
[...]

PR #3034, opened 2026-07-04
Localize profile stats row

punk6529:
## Issue
- The June a11y/i18n workstream localized the profile stats row (stacked PR #2642), but the tail of that PR stack never reached `main` — the `user.statsRow` namespace and messages helper are absent from `main` while all sibling topics landed via the #2604 squash.

## Fix
- Recover the reviewed end-state of that topic from the workstream branch and rebase it onto current `main`.

## Changes
- `components/user/utils/stats/user-stats-row.messages.ts`: new messages helper (`user.statsRow` namespace).
- `components/user/utils/stats/UserStatsRow.tsx`: localized labels and link/button aria-labels for TDH/xTDH/NIC/Rep/followers.
- `i18n/messages/en-US.ts`: adds the `user.statsRow` message block.
- `__tests__/components/user/utils/stats/UserStatsRow.test.tsx`: coverage for the localized labels.

## Validation
- `6529 run typecheck:changed` — passed (50 files).
- `6529 run test -- --testPathPatterns UserStatsRow` — 1 suite, 2 tests passed.

## Risk
- Level: Low
- Why: display-string refactor on one component; no data-shape or routing changes; en-US only (other locales fall back), matching how sibling namespaces landed.
- Rollback: revert the single commit.

## Review Notes
- Helper follows the same `Extract<MessageKey, ...>` pattern as the landed sibling helpers. Main later threaded an optional locale param through some sibling helpers; kept the reviewed simpler form here — happy to align if preferred.


punk6529:
Addressed the i18n review in the latest commit:
- **Locale pinning**: `getUserStatsRowMessage`/`formatUserStatsRowInteger` now take a `SupportedLocale` parameter (default `DEFAULT_LOCALE`), and `UserStatsRow` accepts an optional `locale` prop threaded through every message and number call — same pattern as the landed sibling helpers.
- **Mixed number formatting**: TDH/xTDH/NIC/Rep values now go through a new locale-aware `formatUserStatsRowStatFloor` (floor + `formatInteger`) instead of the non-locale-aware `formatStatFloor`, so all values in the row follow one locale convention.
- **Plural keys**: follower label keys renamed to CLDR `.one`/`.other`.

punk6529:
Follow-up review response:
- **`formatStatFloor` abbreviation claim — deferred as factually incorrect.** `formatStatFloor` (`helpers/Helpers.ts:163-173`) has never abbreviated: it floors and delegates to `formatNumberWithCommas`, producing full grouped integers (`1,234,567`). The abbreviating helper is the unrelated `formatLargeNumber`. For en-US the new `formatUserStatsRowStatFloor` output is character-identical to base behavior (floor → group), so there is no display change and no layout widening.
- **Tests**: added a `followersCount={null}` case (renders as 0) and a `locale="de-DE"` case asserting German grouping (`1.234`) in both the visible value and the aria-label.
- **Locale wiring gap**: acknowledged — callers currently rely on the default, same as the landed sibling components; runtime-locale threading across profile surfaces is tracked workstream-wide rather than per-component.

PR #2645, opened 2026-06-13
Localize profile about edit form

punk6529:
## Summary
- Add message-backed source-locale copy for the profile About edit form textarea, placeholder, character count, buttons, success toast, and moderation errors.
- Give the About edit error container `role="alert"` and a specific dismiss control name.
- Record #2644 as bot-happy and track this stacked follow-up in the workstream memory.

## Validation
- `6529 run test -- --runTestsByPath __tests__/components/user/user-page-header/about/UserPageHeaderAboutEdit.test.tsx __tests__/components/user/user-page-header/UserPageHeaderAboutEditError.test.tsx __tests__/i18n/messages.test.ts --silent --verbose=false --coverageReporters=none`
- `6529 exec eslint components/user/user-page-header/about/UserPageHeaderAboutEdit.tsx components/user/user-page-header/about/UserPageHeaderAboutEditError.tsx i18n/messages/en-US.ts`
- `6529 run typecheck:changed`
- `6529 run react-doctor:diff` (still reports unrelated dirty `contexts/EmojiContext.tsx` fetch-in-effect)
- `git diff --check`
- Next MCP `get_errors`
- Browser smoke on `/punk6529`

Review-ready only. Do not merge this page implementation PR yet.

## Summary by CodeRabbit

* **New Features**
  * About statement editing form now supports multiple languages with localized text for labels, placeholders, buttons, character counter, and error messages.

* **Tests**
  * Enhanced test coverage for the About statement form and error handling to verify functionality and accessibility.

PR #2643, opened 2026-06-13
Localize profile header identity controls

punk6529:
## Issue
- Continue the progressive WCAG 2.2 AA and i18n migration on the user profile header.
- The header identity/media area still had hardcoded labels and disabled edit buttons in read-only public views.

## Fix
- Add a profile-header source-locale message helper and route touched labels through it.
- Replace disabled read-only edit buttons for the profile name and profile picture with plain non-interactive containers.
- Format the profile-enabled month/year through the repo i18n date helper.

## Changes
- Added `user.profileHeader.*` messages for profile name edit labels, profile picture alt/edit labels, banner edit labels, and the profile-enabled date line.
- Updated profile name, profile picture, and banner controls to use message-backed accessible names.
- Updated the read-only profile name/PFP wrappers so public profile views do not expose disabled edit controls.
- Removed the PFP wrapper mount-only state that React Doctor flagged; the portal now guards `document` at render time and only opens after client interaction.
- Added focused tests for accessible names, read-only semantics, image alt text, and i18n fallback behavior.
- Updated the durable frontend a11y/i18n workstream memory.

## Validation
- `6529 run test -- --runTestsByPath __tests__/components/user/user-page-header/name/UserPageHeaderName.test.tsx __tests__/components/user/user-page-header/name/UserPageHeaderNameWrapper.test.tsx __tests__/components/user/user-page-header/UserPageHeaderPfp.test.tsx __tests__/components/user/user-page-header/UserPageHeaderPfpWrapper.test.tsx __tests__/components/user/user-page-header/banner/UserPageHeaderBanner.test.tsx __tests__/i18n/messages.test.ts --silent --verbose=false --coverageReporters=none`: passed, 6 suites / 18 tests.
- Targeted ESLint for touched source files: passed.
- `6529 run typecheck:changed`: passed for 114 changed TypeScript files.
- `6529 run react-doctor:diff`: exited 0 at 98/100; remaining diagnostics are the known unrelated `contexts/EmojiContext.tsx` fetch-in-effect issue and a test-only `next/image` mock warning.
- `git diff --check`: passed; only line-ending warnings on unrelated/local dirty files and touched files.
- Next MCP `get_errors` on the live local frontend: no config or session errors.
- Browser smoke on `/punk6529` at desktop and 390px mobile widths verified no read-only profile edit buttons, message-backed profile image alt text, profile-enabled text, and no mobile horizontal overflow.

## Risk
- Level: Low
[...]

PR #3035, opened 2026-07-04
Localize profile header identity and About controls

punk6529:
## Issue
- The June a11y/i18n workstream localized the profile header (name/pfp/banner) and About statement/edit flows as stacked PRs #2643–#2645, but the tail of that PR stack never reached `main`: the `user.profileHeader` namespace and its messages helper are absent from `main` and `1a-staging`, and the header still renders hardcoded strings without accessible edit-control names.

## Fix
- Recover the reviewed end-state of those topics from the workstream branch and merge it three-way onto current `main`, preserving main's later profile-CMS work (website link, toast error details).

## Changes
- `components/user/user-page-header/user-page-header.messages.ts`: new helper (`user.profileHeader` namespace, display-name + month/year formatting).
- Name/Pfp/Banner components + `UserPageHeaderClient`: shared `profileLabel`, localized aria-labels for edit controls, localized "Profile enabled" date.
- About/AboutStatement/AboutEdit/AboutEditError: localized add/edit/expand/collapse actions, textarea label, character count, save/cancel, moderation error copy.
- `i18n/messages/en-US.ts`: adds the `user.profileHeader` block.
- 9 test suites updated/added under `__tests__/components/user/user-page-header/`.

## Validation
- `6529 run typecheck:changed` — passed (50 files).
- `6529 run test -- --testPathPatterns user-page-header` — 17 suites, 53 tests passed (includes main's pre-existing header suites, so no regression to the newer CMS-era header).

## Risk
- Level: Low-Medium
- Why: touches the profile header render path; changes are display strings and aria-label plumbing only, hand-merged against main's current components; en-US only (other locales fall back), matching sibling namespaces.
- Rollback: revert the single commit.

## Review Notes
- `UserPageHeaderAboutEdit.tsx` and `UserPageHeaderClient.tsx` were three-way merged with main's newer versions — worth a close look at the import/label merges there.


punk6529:
Addressed the review findings in the latest commit:
- **Display-name fallback**: restored the original semantics — `handle → display → mainAddress` — by dropping `profile.primary_wallet` from `getUserProfileHeaderDisplayName`; added a wallet-only regression test asserting the lowercased `mainAddress` renders (and the raw-cased wallet does not).
- **Module-scope ERROR_TEXT**: error copy is now resolved at render time (`getErrorText()`), matching the unknown-error branch and removing the import-time i18n coupling.
- **Hardcoded strings (i18n lane)**: the DM-failure toast and the About save-failed fallback are now message-backed (`user.profileHeader.dm.createFailed.*`, `user.profileHeader.aboutEdit.errors.saveFailed`).
- **Pencil-button keyboard reachability**: not changed — it is a real `<button>`, so it is tab-focusable regardless of `pointer-events`, and the existing `group-focus-within:tw-pointer-events-auto`/opacity classes reveal and enable it on keyboard focus. Happy to add a keyboard-traversal test if you want it covered explicitly.
- **PfpWrapper hydration**: per the follow-up review pass, the `typeof document` guard + "use client" resolves the earlier concern; no change.

PR #3036, opened 2026-07-04
Add nightly branch-janitor workflow

punk6529:
## Issue
- The repo accumulated ~1,000 remote branches (2026-07-04 audit); merged branches were never cleaned up. A one-off prune happened today (167 merged branches deleted with a recovery manifest), but without automation the pile regrows.

## Fix
- A nightly `branch-janitor` workflow that deletes remote branches fully merged into `main` once their last commit is older than 7 days.

## Changes
- `.github/workflows/branch-janitor.yml` (new): nightly cron + manual dispatch with a `dry_run` input; never touches `main`, `1a-staging`, `staging-merge-agent-*`, or open-PR heads; caps at 200 deletions per run; logs every deleted branch name+SHA in the run summary so anything can be restored via `git push origin <sha>:refs/heads/<name>`.

## Validation
- YAML parsed with js-yaml.
- Deletion logic mirrors the manual prune script used for today's 167-branch cleanup (verified against the live branch list: protected sets excluded correctly).
- Suggest first running the workflow manually with `dry_run: true` after merge to review candidates.

## Risk
- Level: Low-Medium
- Why: automation with delete permissions, but scope is strictly merged-into-main branches (recoverable by SHA), with protected-name and open-PR exclusions, a per-run cap, and a dry-run mode.
- Rollback: delete the workflow file; restore any branch from the SHAs logged in the run summary.

## Review Notes
- The 7-day age gate uses the branch tip's commit date (not merge date) — merged branches with recent commits are left for later runs.
- Only same-repo PR heads are excluded; fork heads are unaffected by branch deletion.


punk6529:
Addressed the review in the latest commit:
- **Field-shift corruption**: for-each-ref now emits tab-delimited fields (`%09`) and the loop reads with `IFS=$'\t'` — tabs cannot appear in ref names, unlike `|`.
- **Silently-empty open-PR list**: the heads query now fails closed — the raw API call must succeed, and the result is cross-checked against `search/issues` `total_count`; open PRs with an empty heads list aborts the run before any deletion.
- **Numeric epoch guard**: non-numeric/empty epoch fields are skipped instead of killing the run under `set -e`.
- **First-run safety**: scheduled runs stay in DRY RUN until the repo variable `BRANCH_JANITOR_ENABLED` is set to `true`, so the first nightly run produces a reviewable candidate list with zero deletions. Manual dispatch keeps its `dry_run` input.
- **Nice-to-haves**: jq now compares against `env.GITHUB_REPOSITORY` instead of shell interpolation; added `timeout-minutes: 30`; added an explicit `origin/main` ref assert after the fetch.

PR #3045, opened 2026-07-05
Add push-time secret scanning and credential ignore rules

punk6529:
## Issue
- On 2026-07-04 a rescue snapshot pushed a live bot token (`tmp/` JSON with a JWT + refresh token) to origin. Three gaps let it through: `tmp/` and credential-shaped filenames are not gitignored, the secret scanner only runs on pull requests (direct branch pushes are never scanned), and the scanner's patterns would not have matched a raw JWT sitting in a JSON payload anyway.

## Fix
- Close all three: ignore rules for scratch/credential-shaped paths, a push-triggered scan of every branch push using the same repo-local scanner PR CI already trusts, and a JWT-shape detection pattern — plus a full-tree sweep to confirm nothing credential-shaped is currently tracked.

## Changes
- New `.github/workflows/push-secret-scan.yml`: runs on every branch push; scans the pushed range via `ops/scripts/testing-strategy.cjs scan-changed-secrets` with a merge-base fallback for new branches/force pushes; bare checkout, no install, `contents: read`, no secrets; report artifact kept 30 days. Not a required PR check.
- `.gitignore`: `/tmp/` (nothing tracked under it) and repo-wide `*-auth.json`, `*browser-auth*`, `*storage-state*.json`, `*.credentials.json`, `*-session-token*` (verified zero tracked matches).
- `ops/scripts/testing-strategy.cjs`: new `jwt-like-token` pattern (three-part base64url starting `eyJ` — the leaked file's shape; zero JWT-shaped strings tracked today, so no added noise) and a negative lookahead so `named-secret-assignment` stops matching code expressions (zod chains / env plumbing in `config/*.ts`, `next.config.ts`) that made the scanner a latent PR-blocker for anyone touching those lines.
- `__tests__/scripts/testing-strategy.test.ts`: contract test for the JWT pattern (fixture never echoes the token back).
- `__tests__/playwright/artifactRedaction.test.ts`: bearer-token fixtures now use the repo's existing split-string convention so the scanner stops flagging this test source; runtime strings unchanged.
- `ops/workstreams/repo-health-2026-07/run-log.md`: consolidated Thread A completion + secret-hygiene entry.

## Validation
- Full-tree sweep with the scanner itself over all 6,556 tracked files: 6 findings before this PR (all false positives: 4 config/zod code expressions, 2 fake test fixtures), 0 findings after.
- `test:no-coverage` on the testing-strategy contract suite + artifact-redaction suite: 40/40 green.
- Repo workflow-security validator on the new workflow: zero findings. Simulated the push-scan invocation locally against a real branch range: clean exit.
- Not tested: an actual force-push/new-branch event through the workflow's fallback path (exercised logically and via local merge-base simulation; the first live pushes will confirm).

## Risk
- Level: Low-Medium
- Why: touches the reviewbot-contract scanner (pattern additions are covered by the contract suite) and adds a repo-wide push workflow (read-only, install-free, ~30–60s). Worst case is a false-positive scan failure on a push, which blocks nothing (not a required check) and is visible in the run summary.
- Rollback: revert the commit; the workflow disappears with the file.

## Review Notes
- The named-secret lookahead list (`z.`, `process.`, `publicEnv.`, `privateEnv.`, `serverEnv.`, `import.`, `env.`) is intentionally explicit rather than clever; the sweep proves it clears every current false positive without loosening real-value matches.
- gitleaks was considered for the push scan and rejected: the repo-local scanner is already the PR-time source of truth (reviewbot contract), needs no third-party action or rules file, and runs in seconds on a bare checkout.

### DIFF
--- a/.github/workflows/push-secret-scan.yml
+++ b/.github/workflows/push-secret-scan.yml
@@ -79,24 +79,44 @@ jobs:
 
       # Findings carry file/line/pattern only — the scanner never emits the
       # matched value — so summarizing them here cannot leak the secret.
+      # The guard keeps this step quiet when the job failed for a
+      # non-finding reason (no report file, or a report with no findings).
       - name: Summarize findings on failure
         if: failure()
         run: |
           set -euo pipefail
-          {
-            echo "## Push secret scan: FINDINGS"
-            echo ""
-            echo "A pushed change matched a credential pattern. Rotate the"
-            echo "credential if real, rewrite the branch history to remove it,"
-            echo "then force-push the cleaned branch."
-            echo ""
-            node -e '
-              const report = require("./test-results/push-secret-scan.json");
-              for (const f of report.findings) {
-                console.log(`- ${f.file}:${f.line} (${f.pattern})`);
-              }
-            '
-          } >> "$GITHUB_STEP_SUMMARY"
+          node -e '
+            const fs = require("node:fs");
+            let report;
+            try {
+              report = JSON.parse(
+                fs.readFileSync("test-results/push-secret-scan.json", "utf8")
+              );
+            } catch {
+              console.log(
+                "No scan report produced; the failure happened before or outside the scan."
+              );
+              process.exit(0);
+            }
+            if (report.ok || !Array.isArray(report.findings) || report.findings.length === 0) {
+              console.log("Scan report has no findings; the failure came from another step.");
+              process.exit(0);
+            }
+            const lines = [
+              "## Push secret scan: FINDINGS",
+              "",
+              "A pushed change matched a credential pattern. Rotate the",
+              "credential if real, rewrite the branch history to remove it,",
+              "then force-push the cleaned branch.",
+              "",
+              ...report.findings.map((f) => `- ${f.file}:${f.line} (${f.pattern})`),
+            ];
+            fs.appendFileSync(
+              process.env.GITHUB_STEP_SUMMARY,
+              lines.join("\n") + "\n"
+            );
+            console.log(`Summarized ${report.findings.length} finding(s).`);
+          '
 
       - name: Upload scan report
         if: always()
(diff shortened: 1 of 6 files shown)


punk6529:
Response to the 6529bot general/security reviews (2f66b57a48ba) — fixes pushed in 793b2e549:

**The Important finding (exit-code assertion)**
- `scan-changed-secrets` already fails on findings: `commandScanChangedSecrets` in `ops/scripts/testing-strategy.cjs` sets `process.exitCode = 1` when `result.ok` is false and prints each finding as `error: <file>:<line>: possible secret (<pattern>)`. The code predates this diff, which is why it wasn't visible in the review range. The workflow now documents this at the call site.
- Alert path beyond the artifact: a failing run triggers GitHub's workflow-failure notification to the pusher, and a new `if: failure()` step writes the redacted finding list (file/line/pattern — the scanner never emits matched values) plus rotate/rewrite instructions to the run summary. Push events can't be blocked retroactively by design; detection + loud failure is the goal here.

**Nice-to-haves**
- Lookahead false-negative trade-off: documented in the pattern comment. Accepted as designed — a generated credential value beginning exactly with `z.`/`process.`/`env.` etc. is vanishingly unlikely, and the miss additionally requires the line to name one of the nine watched keys. Anchoring on "value continues as a code expression" costs regex complexity in a reviewbot-contract file for negligible marginal recall.
- JWT segment-length floors: documented in the pattern comment as a deliberate noise/coverage trade-off (defense-in-depth net, not exhaustive JWT detection).
- Ignored-but-committed files: confirmed covered. The workflow checks out the pushed commit, so any committed file — gitignored or not — exists in the working tree; `scan-changed-secrets` derives the file list from `git diff` (not `git status`) and reads the checked-out contents. `.gitignore` only reduces accidental *staging*; it does not hide committed content from the scan.

PR #3052, opened 2026-07-05
Burn down any_casts 358 -> 65: type the NextGen contract layer, platform utils, and component scatter

punk6529:
## Issue
- The debt ratchet's `any_casts` metric sat at 358 (`: any` / `as any` in shippable TS source). Most sites cluster around one root cause: `NextGenContract.abi: any` poisoning every wagmi read downstream, plus scattered untyped error handling, event handlers, and API decode sites. Part of the repo-health dead-patterns workstream.

## Fix
Six commits, root cause first:
1. **NextGen contract layer** — annotate the three NextGen ABIs with viem's `Abi` (house pattern already used by `MEMES_MANIFOLD_PROXY_ABI`), type `NextGenContract.abi`, type every read hook and callback in `nextgen_helpers.ts` precisely (PhaseTimes/AdditionalData/Info/LibraryScript/MintingDetails), then drop ~90 now-redundant `(x.data as any) === true|false` casts across the admin components.
2. **Remaining nextGen sites (51 -> 0)** — event handlers via inference, tuple decodes as `readonly unknown[]` with behavior-identical coercions, `DELEGATION_ABI` casts narrowed `any -> Abi`, precise local shapes (CSSProperties/ReactNode/SwiperClass).
3. **Platform layer** — helpers (`areEqualAddresses`, `isEmptyObject`, fullscreen vendor members), idb-keyval, error-sanitizer, upload-error, drop-hasher, wave-eligibility callbacks, useManifoldClaim decode record, entities (`NFTTokenMetadata` for NFT/Rememe metadata blobs, `DBResponse.next: unknown`).
4. **Timeline** — `any` params were already-declared `string` on the NFTHistory entity; propagated.
5. **Component/app scatter (44 files)** — nft-transfer, lexical edit reconstruction, manifold minting, mapping tools (typed csv-parser imports), distribution plan, pagination, open-graph/pepe services, user pages, model-viewer declaration.
6. **no-base-to-string cleanup** — the sweep's `String(unknown)` coercions tripped the lint rule (blind to the old any-typed versions doing the same); replaced with typeof narrowing / bigint assertions for uint256 reads.

## Changes
- `any_casts` ratchet baseline: **358 -> 65**. The remainder is fully accounted in `ops/workstreams/repo-health-2026-07/any-exceptions.md`: 3 documented exceptions (wagmi `createConnector` conditional return), 1 regex false positive (blog prose "as any"), ~60 deferred to the delegation rewrite (Thread C), 1 deferred to the `src/` fold (Thread D).
- One latent type bug surfaced and fixed: `LibraryScript.script` tuple slot was silently `any`, actually `string[]`.
- Intentional micro-fixes (disclosed): `fetchCollection` returns null instead of crashing when the nextgen collection fetch failed; a few unguarded error-crash paths now degrade to defaults — no happy-path behavior changes anywhere.
- Run-log + exceptions ledger under `ops/workstreams/repo-health-2026-07/`.

## Validation
- Full jest suite green on this branch (1959+ suites; targeted re-runs after every stage: 72/354 nextGen, 138/1486 platform, 1163/5723 scatter, 68/841 post-lint-fix)
- `tsc` (strict, incl. noUncheckedIndexedAccess/exactOptionalPropertyTypes/noPropertyAccessFromIndexSignature): green
- eslint over all 140 changed files: green; knip: green; debt ratchet: green with lowered baseline
- No runtime edits except the disclosed behavior-identical coercions and micro-fixes above

## Risk
- Level: Medium (large surface) mitigated by mechanical patterns
- Why: types-only by policy; every stage validated independently; contract-read coercions (`String`/`Number`/`bigint.toString`) are runtime-identical for viem's decoded values
- Rollback: revert the merge commit

## Review Notes
- `components/delegation/**` intentionally untouched (owned by the styling workstream's rewrite).
- `NextGenAdminUpdateCollection.tsx` keeps a deliberate `i as unknown as number` string-key array lookup to preserve exact legacy semantics for non-canonical numeric input; commented in the diff discussion if needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability across login, access, restricted, and upload flows by handling response and error data more safely.
  * Fixed several cases where missing or empty data could cause incorrect loading, display, or navigation behavior.

* **Refactor**
  * Strengthened type handling across many app areas for more consistent behavior and fewer runtime issues.
  * Updated wallet, leaderboard, NFT, NextGen, and wave-related flows for safer data parsing and rendering.

* **Style**
  * Refreshed some form, table, picker, and list layouts for cleaner spacing and input behavior.

### DIFF
--- a/abis/abis.ts
+++ b/abis/abis.ts
@@ -719,7 +719,7 @@ export const DELEGATION_ABI = [
   },
 ];
 
-export const NEXTGEN_CORE_ABI = [
+export const NEXTGEN_CORE_ABI: Abi = [
   {
     inputs: [
       {
@@ -2278,7 +2278,7 @@ export const NEXTGEN_CORE_ABI = [
   },
 ];
 
-export const NEXTGEN_MINTER_ABI = [
+export const NEXTGEN_MINTER_ABI: Abi = [
   {
     inputs: [
       { internalType: "address", name: "_gencore", type: "address" },
@@ -2732,7 +2732,7 @@ export const NEXTGEN_MINTER_ABI = [
   },
 ];
 
-export const NEXTGEN_ADMIN_ABI = [
+export const NEXTGEN_ADMIN_ABI: Abi = [
   { inputs: [], stateMutability: "nonpayable", type: "constructor" },
   {
     anonymous: false,
--- a/app/access/page.client.tsx
+++ b/app/access/page.client.tsx
@@ -21,9 +21,11 @@ export default function AccessPage() {
       const apiAuth = getStagingAuth();
       fetch(`${publicEnv.API_ENDPOINT}/api/`, {
         headers: apiAuth ? { "x-6529-auth": apiAuth } : {},
-      }).then((r: any) => {
-        r.json().then((response: any) => {
-          setImage(response.image);
+      }).then((r: Response) => {
+        r.json().then((response: { image?: string | undefined }) => {
+          if (response.image) {
+            setImage(response.image);
+          }
         });
         if (r.status !== 401) {
           router.push("/");
@@ -38,7 +40,7 @@ export default function AccessPage() {
     const pass = target.value;
     fetch(`${publicEnv.API_ENDPOINT}/api/`, {
       headers: { "x-6529-auth": pass },
-    }).then((r: any) => {
+    }).then((r: Response) => {
       if (r.status === 401) {
         alert("Access Denied!");
       } else {
--- a/app/api/open-graph/compound/service.ts
+++ b/app/api/open-graph/compound/service.ts
@@ -67,7 +67,9 @@ type CompoundTarget =
 const TX_HASH_LENGTH = 66;
 
 function isPossibleTxHash(value: string): value is Hash {
-  return value.startsWith("0x") && value.length === TX_HASH_LENGTH && isHex(value);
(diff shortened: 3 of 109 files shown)


punk6529:
Addressed the review in bf9846521:
- Guarded both read effects in NextGenAdminAcceptAddressesAndPercentages (SetSplits pattern)
- NextGenAdminUpdateCollection now uses `Number(i)` for bounds check and index — the double-cast is gone, and the latent `[undefined]` selection for non-canonical numeric input with it
- `getError` in the consolidated-addresses item narrows to string; no literal "undefined" toast
- error-sanitizer message extraction typeof-narrows instead of casting
- Ledger: false positive moved to its own section; run-log count filled in (65)
- Reverted the `generated/models/ApiIdentityMuteState.ts` drift (local codegen hook artifact; generated/ stays untouched)

**Deferred:** `mint_cost`/`end_mint_cost` both reading `d[1]` in `useCollectionCosts` is pre-existing decode semantics that this typing PR deliberately preserves (`d[0]` may well be the intended mint cost, but changing which tuple slot feeds the minting UI is a behavior change that needs its own verified fix against the contract ABI — flagged for follow-up rather than smuggled in here).

PR #3066, opened 2026-07-05
feat(waves): auto-focus emoji picker search input on open

brookr:
## Issue
- Opening an emoji picker required a second click into its Search field before typing. Reported for the drop-reaction flow ("Add Reaction" from the hover toolbar or mobile long-press menu) and equally true for the composer's emoji button: by the time the picker is open, focus has already left the editor, so there is nothing to preserve — the extra click is pure friction for keyboard users.

## Fix
- Pass emoji-mart's native `autoFocus` prop (which focuses its search input on mount) through `LazyEmojiPicker`, and enable it at every picker entry point: reaction picker (desktop portal + mobile dialog) and composer picker (desktop portal + mobile dialog). One shared component change covers web, the Electron desktop shell, and the Capacitor mobile wrapper, since all render this same frontend.

## Changes
- `components/waves/LazyEmojiPicker.tsx`: optional `autoFocus` prop forwarded to the emoji-mart `Picker`.
- `components/waves/drops/WaveDropActionsAddReaction.tsx`: `autoFocus` on both desktop and mobile picker instances.
- `components/waves/CreateDropEmojiPicker.tsx`: `autoFocus` on both desktop and mobile picker instances.
- Tests: emoji-mart mocks expose `autoFocus`; new/extended assertions for all four picker instances.

## Validation
- `6529 run test` on both affected suites: 19/19 pass.
- `6529 run lint:quiet`: clean.
- Live Playwright checks against the local dev stack (desktop viewport, iPhone emulation with synthesized long-press, and composer flow): in all three, the deep active element (through emoji-mart's shadow DOM) is the search input immediately after the picker opens, and typing lands in it with no extra click. Screenshots captured.
- Not tested on physical iOS hardware: whether the software keyboard auto-raises on programmatic focus depends on the WKWebView user-action policy; the input is focused either way and behavior is unchanged from before for touch users who tap the field.

## Risk
- Level: Low
- Why: additive optional prop; default behavior unchanged for any other `LazyEmojiPicker` consumer; no data or API impact. Reversible by dropping the prop.

## Deployment
- Frontend only; no backend dependency or ordering constraints.

## Review Notes
- The prop is intentionally opt-in per call site rather than defaulted in `LazyEmojiPicker`, so future embedded uses can decide their own focus behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Emoji pickers now open with the search/input area automatically focused on both desktop and mobile, making reactions and emoji selection faster.
* **Tests**
  * Added coverage to verify the picker focus behavior in create-drop and add-reaction flows.


brookr:
@6529bot review all

(Re-posting the trigger — the previous command got Git-Bash path-mangled to `C:/Program Files/Git/6529bot ...`, which the bot's parser doesn't match.)

punk6529:
/6529bot review all

brookr:
Re the mobile auto-focus nice-to-have: intentional. The picker's primary task is search (the quick-reaction row already serves the no-search case), and the on-screen keyboard appearing immediately matches that intent — same trade-off messaging apps make for their emoji search. Product direction was explicitly to auto-focus on both desktop and mobile for hands-on-keyboard flow, so keeping the paths consistent rather than gating by device.

punk6529:
/6529bot review responsiveness glm-swarm

brookr:
Re the GLM swarm advisories:

- **Mobile soft-keyboard**: intentional product direction (parity with desktop, hands-on-keyboard flow) — see rationale above. Verified in mobile-viewport emulation: the bottom-sheet lays out correctly with the search focused; typing filters immediately.
- **Desktop portal dialog role / focus trap**: pre-existing overlay semantics — this diff only adds the focus behavior, which the dedicated WCAG 2.2 AA review (ran twice on this head) explicitly endorsed as correct for a deliberately-opened picker. Agree the portal could gain `role=dialog` + a trap; that's a separate, wider change — happy to follow up in its own PR.
- **Opt-in vs defaulted prop**: deliberate (noted in the PR body) so future embedded uses choose their own focus behavior.
- **Real-focus assertion**: covered outside jsdom — live Playwright runs assert `document.activeElement` (through the shadow DOM) is the search input on desktop, mobile long-press, and composer flows. Escape-close is already in the unit suite.

punk6529 (approved):
Last-pass review: clean, focused change. Additive optional `autoFocus` prop forwarded through LazyEmojiPicker to emoji-mart's Picker, wired into all four picker entry points (reaction + composer, desktop + mobile) with matching test coverage. All CI checks green, CodeRabbit had no actionable comments, no unresolved threads. LGTM.

PR #3071, opened 2026-07-05
Run staging E2E packs automatically after staging deploys

punk6529:
Closes the gap found during the Redux deploy train: staging is access-gated, the Playwright staging packs existed but nothing ran them, and the access code lived only on one machine.

## What
- `PLAYWRIGHT_STAGING_ACCESS_CODE` added as a repo secret (value validated against the staging gate — HTTP 200 on a gated route — before storing; the e2e harness already redacts it from artifacts).
- New `Staging E2E` workflow: triggers on successful **Web Deploy - STAGING** runs (+ manual dispatch), checks out the deployed SHA, installs via the standard Socket-Firewall/pnpm path, and runs all 12 staging packs with a per-pack step summary. Runs every pack even after a failure; uploads Playwright artifacts on failure.
- Least-privilege: `contents: read`, no credential persistence, actions pinned by SHA (mirrors app-pr-ci).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added an automated staging end-to-end test run that triggers after successful staging deployments and can also be started manually.
  * Enhanced test reporting with clear per-pack success/failure summaries and improved failure details.
* **CI/CD**
  * When staging E2E fails, the test results and Playwright report are uploaded as artifacts for easier troubleshooting (retained for 7 days).

PR #3077, opened 2026-07-05
Keep explicitly set page titles from being clobbered by App Router head commits

punk6529:
## Problem

The post-deploy **Staging E2E** workflow (added today in #3071) has never gone green: `Network Metrics downloads route reads public upload metadata` fails `toHaveTitle("Network Metrics | Open Data")` — **against production too** — and `renders The Memes mint page read-only` fails on staging.

Root cause (traced with a `<title>` MutationObserver on a live page): the App Router commits the route's server metadata `<title>` after hydration settles, and that commit **overwrites** `document.title` values already written from `TitleContext`:

```
"Network Metrics"                    <- SSR metadata
"Network Metrics | Open Data"        <- useSetTitle enrichment applies
"Network Metrics"                    <- head commit clobbers it back (final)
```

Pages that set their title on mount (`useSetTitle`) lose deterministically; pages that set it after a data fetch (The Memes mint) win or lose per environment timing — which is exactly the staging-vs-production split the packs showed.

## Fix

- `TitleContext` now tracks **ownership**: titles claimed via `setTitle`/wave data own `document.title`; the provider's route-default placeholders do not (server metadata must keep winning there, e.g. `/open-data` → "Open Data").
- `DynamicHeadTitle` re-asserts owned titles via a `MutationObserver` on `document.head` (covers both text overwrites and `<title>` node replacement) and releases ownership on route change. Unowned titles get at most a one-shot fallback write when `document.title` is empty.

## Verification

- Live dev serve: `/open-data/network-metrics` → **"Network Metrics | Open Data"** (previously clobbered), `/open-data` → "Open Data", `/network` → "Network" — all matching the staging pack assertions.
- New unit tests: owned-title re-assertion, unowned passthrough (metadata wins), empty-title one-shot fallback, root-path default, unmount cleanup, and provider ownership lifecycle across route changes.
- `typecheck:ci` and eslint clean.

## Summary by CodeRabbit

* **New Features**
  * Added title ownership tracking to preserve the browser tab title correctly across route changes, wave-derived titles, and metadata commits.
  * Exposed `isTitleOwned` and `titlePathname` to improve route-aware title behavior.

* **Bug Fixes**
  * Improved dynamic title handling to prevent unwanted overwrites when unowned, while still respecting server metadata and root-path defaults.
  * Added resilient reassertion when the `<title>` element is replaced.

* **Tests**
  * Expanded Jest/RTL coverage for ownership transitions, release semantics across renders/pathnames, DOM replacement, and cleanup.

### DIFF
--- a/__tests__/components/dynamic-head/DynamicHeadTitle.test.tsx
+++ b/__tests__/components/dynamic-head/DynamicHeadTitle.test.tsx
@@ -1,0 +1,99 @@
+import { render, waitFor } from "@testing-library/react";
+import React from "react";
+
+let mockTitle = "6529.io";
+let mockIsTitleOwned = false;
+let mockPathname = "/network";
+
+jest.mock("@/contexts/TitleContext", () => ({
+  DEFAULT_TITLE: "6529.io",
+  useTitle: () => ({ title: mockTitle, isTitleOwned: mockIsTitleOwned }),
+}));
+
+jest.mock("next/navigation", () => ({
+  usePathname: () => mockPathname,
+}));
+
+import DynamicHeadTitle from "@/components/dynamic-head/DynamicHeadTitle";
+
+const flushObservers = async () => {
+  await new Promise((resolve) => setTimeout(resolve, 0));
+};
+
+describe("DynamicHeadTitle", () => {
+  beforeEach(() => {
+    mockTitle = "6529.io";
+    mockIsTitleOwned = false;
+    mockPathname = "/network";
+  });
+
+  it("re-asserts an owned title after a later head commit overwrites it", async () => {
+    mockTitle = "Network Metrics | Open Data";
+    mockIsTitleOwned = true;
+    document.title = "Network Metrics"; // SSR metadata title
+
+    render(<DynamicHeadTitle />);
+    expect(document.title).toBe("Network Metrics | Open Data");
+
+    // The App Router's metadata commit lands after hydration and would
+    // silently clobber the owned title without the observer.
+    document.title = "Network Metrics";
+    await waitFor(() =>
+      expect(document.title).toBe("Network Metrics | Open Data")
+    );
+  });
+
+  it("lets server metadata win over route-default placeholders", async () => {
+    mockTitle = "Open Data | Tools"; // provider route default, not owned
+    mockIsTitleOwned = false;
+    document.title = "Open Data";
+
+    render(<DynamicHeadTitle />);
+    expect(document.title).toBe("Open Data");
+
+    document.title = "Open Data Updated";
+    await flushObservers();
+    expect(document.title).toBe("Open Data Updated");
+  });
+
+  it("writes a one-shot fallback when the document title is empty", async () => {
+    mockTitle = "Open Data | Tools";
+    mockIsTitleOwned = false;
+    document.title = "";
+
+    render(<DynamicHeadTitle />);
+    expect(document.title).toBe("Open Data | Tools");
+
+    // Not sticky: a later head commit must win for unowned titles.
+    document.title = "Open Data";
+    await flushObservers();
+    expect(document.title).toBe("Open Data");
+  });
+
+  it("owns the default title on the root path", async () => {
+    mockTitle = "6529.io";
+    mockIsTitleOwned = false;
+    mockPathname = "/";
+    document.title = "Something Else";
+
+    render(<DynamicHeadTitle />);
+    expect(document.title).toBe("6529.io");
+
+    document.title = "Something Else";
+    await waitFor(() => expect(document.title).toBe("6529.io"));
+  });
+
+  it("stops re-asserting after unmount", async () => {
+    mockTitle = "Network Metrics | Open Data";
+    mockIsTitleOwned = true;
+    document.title = "Network Metrics";
+
+    const { unmount } = render(<DynamicHeadTitle />);
+    expect(document.title).toBe("Network Metrics | Open Data");
+
+    unmount();
+    document.title = "Network Metrics";
+    await flushObservers();
+    expect(document.title).toBe("Network Metrics");
+  });
+});
--- a/__tests__/components/dynamic-head/DynamicHeadTitle.test.tsx
+++ b/__tests__/components/dynamic-head/DynamicHeadTitle.test.tsx
@@ -3,11 +3,16 @@ import React from "react";
 
 let mockTitle = "6529.io";
 let mockIsTitleOwned = false;
+let mockTitlePathname: string | null = "/network";
 let mockPathname = "/network";
 
 jest.mock("@/contexts/TitleContext", () => ({
   DEFAULT_TITLE: "6529.io",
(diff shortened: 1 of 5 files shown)


punk6529:
@6529bot Round-1 findings addressed in 177713f4a (plus the `Installed app checks` failure in `TitleContext.test.tsx`):

**Important — navigation fight between owned-title routes: eliminated structurally.** Ownership is now keyed to the pathname it was claimed for (`explicitTitlePathname === pathname`), so it evaporates in the *same render* as a client-side navigation — before the App Router commits the next route's metadata and before any effect runs. The observer is disconnected in that same commit's cleanup, so there is no window where a stale owned title can fight the framework. Covered by the new "releases the title to the next route's metadata on navigation" test.

**CI failure (existing `TitleContext.test.tsx`)** — those tests encode a real contract the round-1 ownership model missed: leaving a wave *without a pathname change* (e.g. dropping `?wave=` on `/messages`) triggers no metadata commit, so the context's route-default reset must still reach `document.title`. Unowned titles now get a **write-on-change** (never on initial load, no observer): defaults are restored on transitions, while server metadata still wins fresh loads and pathname navigations. Full suite passes again (19 tests across the three title suites).

**Nice-to-haves — all fixed**
- Observer scoped to the `<title>` node's text; a shallow (`childList` only, non-subtree) head observer re-arms it if a commit replaces the node — head churn from meta/script injection no longer triggers the callback. Node replacement is covered by a new test.
- Null-`document.head` guard added.
- Negative-assertion tests now flush microtasks plus two macrotask ticks instead of a single `setTimeout(0)`.

Live re-verified on a dev serve: `/open-data/network-metrics` → "Network Metrics | Open Data" (sticky through head commits), `/open-data` → "Open Data", `/network` → "Network".

punk6529:
@6529bot Round-2 findings addressed in eab39104f:

**Important — cross-route clobber risk: eliminated, with the ordering verified live.** The provider now exposes `titlePathname` (the pathname the current title text was computed for), and the unowned write-on-change fires **only** for same-pathname context transitions with consistent context (`titlePathname === pathname`, previous observation on the same pathname). Across a route change the component never writes, so the new route's metadata commit wins regardless of whether it lands before or after the effect. Coverage:

- New unit test "never writes route defaults across a route change, whichever side of the metadata commit" stages exactly the ordering you flagged: nav render (context inconsistent → untouched), metadata commit, provider reset render (consistent but cross-route → no write) — metadata stands.
- Live browser check on a dev serve: owned `/open-data/network-metrics` ("Network Metrics | Open Data") → real client-side `<a>` navigation to unowned `/tools/api` → final title is the metadata "API", not a placeholder, not the stale owned title.
- The same-pathname restore contract (wave exit on `/messages`) still holds — the pre-existing `TitleContext.test.tsx` suite passes, with its meme-calendar case updated to simulate the metadata commit jsdom lacks (that route's metadata title equals the old expected string, so the end state is unchanged; the test now also proves the context doesn't fight the commit).

**Nice-to-haves — both added**
- Mounted-setter variant: the setter survives the pathname change; the test proves the provider's reset effect clears the old claim and the setter's effect re-claims for the new route (`title-pathname` reporter asserts the re-claimed pathname).
- Wave-data ownership branch covered: `/messages?wave=wave-1` + `useSetWaveData` → `isTitleOwned` true with "Wave One | Brain".

21 tests pass across the three title suites; eslint and typecheck clean.

punk6529:
Closing after ground-truthing against deployed reality — the runtime change here is the wrong fix.

Production today is self-consistent: **late, data-driven `useSetTitle` calls win** the title race (The Memes mint page ends enriched — the media pack passes on prod), while **mount-time calls lose to the App Router's metadata commit** (`/the-memes`, `/about/*`, `/open-data/network-metrics` all show their server metadata titles — exactly what the always-green PR-CI smoke packs assert). This PR made mount-time explicit titles win, which flipped user-visible titles on established pages and broke the smoke pack (`The Memes` → `The Memes | Collections`, `Subscriptions` → `Subscription Minting | About`, etc.) — the round-3 `Installed app checks` failure.

The two spec families in the repo froze **opposite sides of the same race** (smoke packs = metadata side; the new staging packs' two title assertions = enrichment side, authored against local-dev timing where the race resolves the other way). Which side *should* win on which pages is a product decision that shouldn't be made inside a deploy train.

Follow-ups:
- The never-green staging pack assertions get aligned to deployed reality in a small spec-only PR (metadata title for `/open-data/network-metrics`; timing-tolerant mint title).
- The underlying title-semantics contradiction (mount-time `useSetTitle` is a no-op on full loads; several route defaults are never user-visible) is filed for a deliberate decision, with this PR's analysis and test corpus as the starting point.

punk6529:
Heads-up from the app-wide title audit: #3089 aligns every static route's generateMetadata title with its client-intended string (section suffixes included) and updates the Playwright/jest expectations to those canonical values.

Suggested merge order: #3089 first, then this PR. Reason: with mismatched strings still on main, ownership re-assertion flips e2e expectations on deployed envs (e.g. tests/pages/the-memes.spec.ts and the smoke pack assert "The Memes" today, but an owned client title yields "The Memes | Collections"). After #3089 the static strings are identical on both sides, so this PR only changes behavior for genuinely dynamic titles (wave unread counts, mint/NFT detail, notification counts) — exactly the cases it protects.

One nit in the PR body: /open-data/network-metrics client-side sets "Consolidated Network Metrics | Open Data" since #3081 (VIEW enum no longer serializes as undefined across the RSC boundary), so the pack now asserts the Consolidated string rather than "Network Metrics | Open Data".

punk6529:
Reopening — the closing objection was ordering-dependent, and the ordering problem is now resolved.

The closure (2026-07-05 14:22Z, before #3089 existed) correctly observed that merging this PR alone flips user-visible titles on established pages and breaks the smoke packs, and asked for a deliberate product decision on which side of the race should win. That decision has since been made by the release owner: the client-intended titles (section suffixes, dynamic enrichment) are the canonical ones. #3089 — now on `main` — implements the static half app-wide: every route's `generateMetadata` emits the client-intended string, and all Playwright/jest title expectations assert those canonical values.

With #3089 merged, this PR no longer changes any user-visible static title (SSR and client strings are identical, so re-assertion is a no-op there). What it does change is exactly the dynamic set it was written for: mount-and-fetch titles (`Mint #N | <name> | The Memes`, NFT/profile detail), wave unread counts, and notification counts become deterministic on deployed builds instead of racing the metadata commit — the same enrichment production already resolves to when the timing happens to fall that way, and the state #3088's tolerance treats as the better outcome.

Also pushed alongside the `main` sync: a small hardening commit. MutationObserver callbacks are microtasks and can fire between an SPA navigation's head commit and this effect's cleanup, letting a stale observer re-assert the previous route's title over the new route's metadata; the observer now captures the location pathname when armed and bails once it moves (with a regression test that fails without the guard).

Ordering context: merged after #3089 (static alignment) and alongside #3088 (tolerant mint gate, which continues to pass with the now-deterministic enriched title).

Issue #3078, opened 2026-07-05
Delegation use-case lock UI reads multicall envelopes instead of .result


punk6529:
## What

`components/delegation/CollectionDelegation.tsx` reads the use-case lock statuses through `useReadContracts` (`useCaseLockStatuses`, `useCaseLockStatusesGlobal`). With wagmi v2 multicall (default `allowFailure: true`), `data[i]` is the envelope object `{ error?, result?, status }` — the decoded boolean lives at `data[i].result`. Several branches treat the envelope itself as the boolean:

- the asterisk marker: `useCaseLockStatusesGlobal.data?.[index] ? " *" : ""` — an envelope object is always truthy, so the asterisk shows for every use case once data loads
- the LOCKED/UNLOCKED select labels: `useCaseLockStatuses.data?.[index] || useCaseLockStatusesGlobal.data?.[index] || collectionLockRead.data ? " - LOCKED" : " - UNLOCKED"` — same always-truthy problem, so every use case renders LOCKED
- the lock/unlock button branch compares the envelope with `false` (`(data[lockUseCaseIndex] as unknown as boolean) === false`), which can never be true for an object, so the branch degenerates to a data-presence check
- the in-flight toast title picks `"Unlocking"` whenever the envelope exists: `useCaseLockStatuses?.data?.[lockUseCaseIndex] ? "Unlocking" : "Locking"`

The same file shows the correct pattern elsewhere: `retrieveIncomingConsolidations.data?.[index]?.result`.

The single-call `useReadContract` reads (`collectionLockRead`, `collectionLockReadGlobal`) are fine — their `data` is the decoded boolean.

## Why it is filed instead of fixed

Found while typing the delegation area during the repo-health campaign (any_casts burn-down). The typing PRs are strictly behavior-preserving, and correcting these reads changes visible UI behavior in a wallet flow, so it needs its own change with real verification (likely: read `.result`, then re-check the asterisk/LOCKED semantics against the contract's `retrieveCollectionUseCaseLockStatus`).

## Where

- `components/delegation/CollectionDelegation.tsx` — search for `useCaseLockStatuses`, `useCaseLockStatusesGlobal`, `lockUseCaseIndex`.

PR #3072, opened 2026-07-05
Add Copy text option to wave drop mobile menu

punk6529:
## Problem

On touch devices, text selection in waves/DMs is intentionally disabled (`touch-select-none` / `touch-action-sheet-select-none`) so that long-press opens the drop action sheet instead of the native selection callout. But the action sheet only offered **Copy link** — there was no way to copy a message's *text* on mobile. Users have resorted to screenshots + OCR to get message content into other tools (reported by @0xcuttlefish).

## Solution

- **New `WaveDropMobileMenuCopyText`** action-sheet item ("Copy text"), shown directly above "Copy link" in both the authenticated and guest menus. Same interaction pattern as Copy Link: `Copied!` confirmation, closes the menu on copy, disabled for temporary (in-flight) drops, clicks don't bubble to the parent card.
- **Extracted the pure formatting logic** (~650 lines: markdown→plain conversion, reply/quote references, embeds, attachment URLs, author + timestamp headings) out of `useWaveDropsClipboard` into a new shared `helpers/waves/drop-clipboard.helpers.ts`, plus a `buildDropClipboardText(drop)` convenience. The hook now contains only the DOM/selection logic and imports the formatting — no behavior change to desktop Ctrl+C copy.
- Mobile "Copy text" therefore produces **the same output as desktop copy**, e.g. `alice (03:46): message text`, including reply/quote context and media URLs.

## Testing

- New: `__tests__/components/waves/drops/WaveDropMobileMenuCopyText.test.tsx` (success/timer flow, attachment URLs, no click bubbling, clipboard failure, temp-drop disable)
- New: `__tests__/helpers/waves/drop-clipboard.helpers.test.ts` (plain/markdown formats, reply and quote context, attachment dedup, winner type/rank annotation)
- Existing `WaveDropMobileMenu` and `WaveDropMobileMenuCopyLink` suites pass unchanged (29 tests across 4 suites)
- `tsc` and ESLint clean on all changed files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a **Copy text** option to the mobile drop menu alongside **Copy link**, with **Copied!** confirmation.
  * Clipboard text now supports **plain** and **markdown** formatting, including author/reply/quotes and attachment links.
* **Localization**
  * Added new **copy**-related translations for multiple languages.
* **Bug Fixes**
  * Copy actions stop propagation, and copying is **disabled for temporary drops**.
* **Tests**
[...]

PR #3079, opened 2026-07-05
Burn down any_casts 64 -> 45: type the delegation write-config path

punk6529:
## Issue
- Repo-health campaign, workstream E tail: `components/delegation/**` carries ~60 of the 64 remaining ratchet-counted `any` sites. They were deferred by the earlier NextGen typing wave (#3052) to an expected delegation rewrite that turned out unnecessary, so they need direct typing. This is slice 1 of 3: the contract-write config path shared by the delegation forms.

## Fix
- Type the seam instead of the symptoms: `DELEGATION_ABI` is annotated `: Abi` at the source (the same treatment #3052 gave the NextGen ABIs), a new `DelegationWriteParams` interface in `delegation-shared.ts` models the write config the form components assemble, and `DelegationSubmitGroups` consumes it instead of `writeParams: any`.

## Changes
- `abis/abis.ts`: `DELEGATION_ABI: Abi`. Existing `DELEGATION_ABI as Abi` casts elsewhere (NextGenMint) become no-ops and are deliberately left untouched — those files belong to a concurrent workstream.
- `components/delegation/delegation-shared.ts`: `DelegationWriteParams`; `getGasError(error: any)` -> `getGasError(error: Error)`.
- `components/delegation/DelegationFormParts.tsx`: `writeParams: DelegationWriteParams`; the submit handler narrows `functionName` before `writeContract` — submission was already gated by the same `validate()` predicate that decides whether `functionName` is set, so the new early-return is not reachable in production flows.
- `components/delegation/DelegationToast.tsx`: `DelegationToastState` exported.
- `NewDelegation.tsx` / `NewAssignPrimaryAddress.tsx`: callback props typed (`void`, `DelegationToastState`); the legacy `onSettled(data: any, error: any)` annotations inside the write configs become `(data: unknown, error: Error | null)`. Note these callbacks are inert under wagmi v2 (mutation callbacks are the second argument of `writeContract`, not part of the variables object); they are typed, not removed, to keep this PR behavior-preserving.
- `__tests__/.../DelegationSubmitGroups.test.tsx`: the valid-submit fixture now carries a `functionName` like every production caller; a new test pins the no-`functionName` no-op contract.
- Ratchet baseline: `any_casts` 64 -> 45. `oversized_files` 90 -> 89 rides along per the stale-baseline protocol (`useWaveDropsClipboard.ts` healed by #3072; `--update` shrank the grandfather list).

## Validation
- `tsc --noEmit -p tsconfig.typecheck.json` green at this branch state.
- Full Jest suite green at this branch state (numbers in PR comment/CI); the delegation suites plus the stacked follow-up slices were also green together locally.
- `lint:changed` clean; `node scripts/debt-ratchet.cjs` passes with the lowered baseline.

## Risk
- Level: Low
- Why: type-only edits plus one unreachable guard in a wallet flow; no runtime object shapes change (the write config object still carries its legacy fields through the spread).
- Rollback: revert the merge commit.

## Review Notes
- The inert-`onSettled` finding and the pre-existing lock-status envelope reads discovered during this work are tracked in issue #3078 rather than fixed here; slices 2 and 3 (follow-up PRs) finish the remaining delegation `any` sites.

## Summary by CodeRabbit

- **Bug Fixes**
  - Delegation submissions now include the selected action when sent, and submissions are skipped when that action isn’t available.
  - Toast and error handling in delegation flows is more consistent, helping prevent unexpected failures.

- **Chores**
  - Improved internal typing and validation around delegation features for better reliability.
  - Updated supporting tests to cover the new submission behavior.

### DIFF
--- a/__tests__/components/delegation/DelegationSubmitGroups.test.tsx
+++ b/__tests__/components/delegation/DelegationSubmitGroups.test.tsx
@@ -57,7 +57,7 @@ describe("DelegationSubmitGroups", () => {
     render(
       <DelegationSubmitGroups
         title="T"
-        writeParams={{ foo: "bar" }}
+        writeParams={{ foo: "bar", functionName: "registerDelegationAddress" }}
         showCancel={true}
         gasError={undefined}
         validate={() => []}
@@ -67,11 +67,32 @@ describe("DelegationSubmitGroups", () => {
     );
     expect(screen.getByText("Cancel")).toBeInTheDocument();
     fireEvent.click(screen.getByRole("button", { name: "Submit" }));
-    expect(mockWriteContract).toHaveBeenCalledWith({ foo: "bar" });
+    expect(mockWriteContract).toHaveBeenCalledWith({
+      foo: "bar",
+      functionName: "registerDelegationAddress",
+    });
     expect(onSetToast).toHaveBeenCalledWith({
       title: "T",
       message: "Confirm in your wallet...",
     });
+  });
+
+  it("does not submit while writeParams lack a functionName", () => {
+    const onSetToast = jest.fn();
+    render(
+      <DelegationSubmitGroups
+        title="T"
+        writeParams={{ foo: "bar", functionName: undefined }}
+        showCancel={false}
+        gasError={undefined}
+        validate={() => []}
+        onHide={jest.fn()}
+        onSetToast={onSetToast}
+      />
+    );
+    fireEvent.click(screen.getByRole("button", { name: "Submit" }));
+    expect(mockWriteContract).not.toHaveBeenCalled();
+    expect(onSetToast).not.toHaveBeenCalled();
   });
 
   it("shows receipt errors instead of success toast", () => {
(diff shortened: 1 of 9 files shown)


punk6529:
GLM advisory dispositions (all defer-with-evidence):

1. **getGasError callers**: its only callers are the six form components' onSettled callbacks; the two touched here type error as `Error | null` and narrow with `if (error)` before calling, the four untouched ones still pass `any` (slice 2 tightens them). Those callbacks are also inert under wagmi v2 — mutation callbacks are writeContract's second argument, not part of variables — so no runtime path reaches getGasError with a non-Error today, and viem errors are Error subclasses regardless.
2. **Test fixtures vs typecheck**: `__tests__/**` is excluded from tsconfig.typecheck.json, and the Installed app checks lane (typecheck:changed) passed on this head — the fixtures are runtime-only by design.
3. **Baseline drop is real**: the Debt ratchet required check re-counts on the PR head and passed (actual 45 vs baseline 45); the 19 removed sites are the 9+8 prop/onSettled annotations in NewDelegation/NewAssignPrimaryAddress plus getGasError and writeParams.

PR #3083, opened 2026-07-05
Staging E2E hardening: token-metadata read-POST allowlist + wallet-checker budget

punk6529:
## Why

This PR started as the full fix for the deterministic post-deploy Staging E2E failures ([run 28741346113](https://github.com/6529-Collections/6529seize-frontend/actions/runs/28741346113)). While it was in review, sibling PRs landed the bulk of it on main:

- NextGen ~12px horizontal overflow → #3080 (restored container-fluid gutters)
- Alchemy `contracts` read-POST guard allowlist → #3086
- Collection titles / `Status:` combobox locator → #3082, #3089 (metadata aligned with client titles, killing the title race at the root)
- Mint-page title tolerance → `66491f9ae`

The branch has been rebuilt on current main and now carries **only the two remaining deltas**:

## 1. Guard: allow the `token-metadata` read-POST twin

`POST /api/alchemy/token-metadata` (and its `api.<env>.6529.io/alchemy-proxy/token-metadata` fallback) is the same POST-as-read batch lookup class as the already-allowlisted `contracts` endpoints — `fetchTokenMetadataFromApi` in `hooks/useAlchemyNftQueries.ts` fires it from NFT-rendering surfaces (token-list, NftPicker) and would trip the read-only guard exactly like the `/punk6529/xtdh` contracts flake did. The route handler is a pure read (batched `getNFTMetadataBatch` proxy, POST only to carry the token list). Unit coverage extended; other POSTs on those hosts stay blocked.

## 2. Wallet Checker empty-state budget

If the "Checking delegation records..." indicator hasn't appeared yet when the `toBeHidden` assertion runs, it passes instantly and the empty-state text gets only the default 5s while the delegation lookup is still in flight — the flake signature documented on 2026-07-05. The empty-state assertion now gets the same 20s budget.

## Verification

- Guard unit tests: 16/16.
- delegation / waves-profile / media / network-open-data packs green against staging.6529.io with these exact deltas (run pre-rebuild; both changes byte-identical).

### DIFF
--- a/__tests__/playwright/readonlyMutationGuard.test.ts
+++ b/__tests__/playwright/readonlyMutationGuard.test.ts
@@ -125,7 +125,9 @@ describe("Playwright read-only mutation guard", () => {
   it("allows first-party read-only alchemy proxy POST lookups", () => {
     for (const url of [
       "https://api.6529.io/alchemy-proxy/contracts",
+      "https://api.6529.io/alchemy-proxy/token-metadata",
       "https://api.staging.6529.io/alchemy-proxy/contracts",
+      "https://api.staging.6529.io/alchemy-proxy/token-metadata",
     ]) {
       expect(
         decideReadonlyRequest({
@@ -171,6 +173,7 @@ describe("Playwright read-only mutation guard", () => {
       "https://6529.io/api/open-graph",
       "https://6529.io/api/twitter/preview",
       "https://6529.io/api/alchemy/contracts",
+      "https://6529.io/api/alchemy/token-metadata",
     ]) {
       expect(
         decideReadonlyRequest({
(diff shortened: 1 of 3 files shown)


punk6529:
Follow-up on the lane verdicts (all posted for the pre-rebuild head `ddb3c69`): the branch has since been rebuilt on current main because #3080/#3082/#3086/#3089 landed the NextGen overflow fix, the title/metadata alignment, and the contracts guard allowlist independently. The head now carries only the two remaining test-side deltas (token-metadata read-POST allowlist + wallet-checker empty-state budget) — a strict subset of what the lanes reviewed and passed.

Addressing the one GLM-swarm advisory that still applies to the current diff:

> If the proxy forwards arbitrary JSON-RPC (e.g., `eth_sendTransaction`), a readonly E2E run could issue a state-changing call through this allowlisted path.

The allowlisted paths are not JSON-RPC forwarders. `app/api/alchemy/token-metadata/route.ts` (like `contracts`) validates a strict request shape (checksummed addresses / token ids only, bounded batch size and body bytes) and issues fixed Alchemy NFT-API REST lookups (`getNFTMetadataBatch` / `getContractMetadata`); there is no method field to smuggle a mutating call through. The guard allowlists exactly these two POST paths per host — every other POST on `api.<env>.6529.io` stays blocked, covered by the extended unit test.

PR #3085, opened 2026-07-05
Localize clipboard timestamps and surface copy failures in wave drop menu

punk6529:
Two polish items deferred from the PR #3072 review threads (wave drop mobile Copy text, shipped in web 4.69.0).

## 1. Locale-aware clipboard timestamps (i18n lane, `i18n/no-direct-locale-formatting`)

- New `formatTime` helper in `i18n/format.ts` (mirrors `formatDate`, but returns an empty string for invalid values so callers can omit the time suffix entirely — preserving the previous clipboard behavior).
- `drop-clipboard.helpers.ts` no longer calls `Intl.DateTimeFormat(undefined, ...)`; the locale is an explicit required parameter of `formatMessage` / `formatMessages` / `buildDropClipboardText` since the module is pure.
- Both consumers resolve the locale via `useBrowserLocale()`: the desktop copy path (`useWaveDropsClipboard`) and the mobile Copy text action. Desktop and mobile output stay identical.

## 2. Copy-failure feedback (general / WCAG / GLM-swarm lanes)

`WaveDropMobileMenuCopyAction` (shared by Copy text and Copy link) previously called `onCopy()` on both the rejected clipboard write and the missing-API fallback, closing the menu as if the copy had succeeded. Now:

- The menu **stays open** on failure so the user can see the state and retry.
- The label flips to a transient red "Copy failed" for 2s (same timing as the "Copied!" state).
- The failure is announced through the existing `tw-sr-only` `role="status"` region, which stays **outside** the label span (Chromium drops live-region content from the button's name-from-contents computation).
- Considered the `setToast` primitive from `AuthContext`, but the action sheet is a full-screen portal layer and the review thread asked for inline label feedback; this also keeps the failure feedback adjacent to the button that triggered it.

## i18n

New key `waves.drop.actions.copyFailed` in `en-US` with ASCII-safe `fr-FR` / `es-ES` / `de-DE` translations; `en-GB` falls back to `en-US` like the sibling keys.

## Tests

- `drop-clipboard.helpers`: locale-formatted headings (en-US vs de-DE, computed via Intl so TZ-independent), invalid-timestamp fallback, updated signatures.
- `WaveDropMobileMenuCopyText`: failure keeps the menu open + shows/announces/resets "Copy failed"; missing-clipboard-API failure; browser-locale (de-DE) end-to-end timestamp case.
- `i18n/messages`: `formatTime` locale/invalid cases and the new key across all locales.

All 20 tests in the three touched suites pass; `typecheck:ci`, `eslint --max-warnings=0` on changed files, and `knip` are clean locally.

## Summary by CodeRabbit

* **New Features**
  * Copying wave drop content now respects the browser’s locale for timestamps and formatted text.
  * A new “Copy failed” message appears when clipboard access is unavailable or a copy attempt fails.

* **Bug Fixes**
  * Copy actions now keep the menu open on failure, allowing users to retry.
  * Timestamp display now handles invalid values gracefully by omitting the time.

* **Tests**
  * Expanded coverage for locale-specific formatting, clipboard failure handling, and translated status messages.

### DIFF
--- a/__tests__/components/waves/drops/WaveDropActionsCopyText.test.tsx
+++ b/__tests__/components/waves/drops/WaveDropActionsCopyText.test.tsx
@@ -45,7 +45,9 @@ describe("WaveDropActionsCopyText", () => {
 
     fireEvent.click(screen.getByRole("button", { name: "Copy text" }));
 
-    expect(writeText).toHaveBeenCalledWith(buildDropClipboardText(drop));
+    expect(writeText).toHaveBeenCalledWith(
+      buildDropClipboardText(drop, "en-US")
+    );
     await waitFor(() => expect(onCopy).toHaveBeenCalled());
   });
 
--- a/__tests__/components/waves/drops/WaveDropMobileMenuCopyLink.test.tsx
+++ b/__tests__/components/waves/drops/WaveDropMobileMenuCopyLink.test.tsx
@@ -128,7 +128,8 @@ describe("WaveDropMobileMenuCopyLink", () => {
     expect(parentClick).not.toHaveBeenCalled();
   });
 
-  it("closes the menu once when clipboard write fails", async () => {
+  it("keeps the menu open and shows failure feedback when clipboard write fails", async () => {
+    jest.useFakeTimers();
     const clipboardWrite = createDeferredClipboardWrite();
     writeText.mockReturnValueOnce(clipboardWrite.promise);
     const onCopy = jest.fn();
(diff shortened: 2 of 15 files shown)


punk6529:
Addressed the lane feedback in 0402fe135:

- **GLM-swarm (important)**: the CopyText tests now capture and restore the original `navigator.clipboard` / `navigator.languages` descriptors instead of leaving overrides behind.
- **GLM-swarm (nice-to-have)**: the missing-clipboard-API test now also asserts the `Copy failed` label and `role=status` region reset after the 2s timeout.
- **general lane**: reworded the missing-API comment — the menu stays open so the failure is visible, not because retry would necessarily succeed in a permanently clipboard-less context.

Skipped with reason:
- **fr-FR accent** (`Échec`): ASCII-safe strings are the deliberate convention for these locale files, as the i18n lane's own review confirmed for this PR.
- **Transient accessible-name change during copied/failed states**: intentional tradeoff documented in the component — the live region must stay outside the label because Chromium drops live-region content from name-from-contents.
- **Responsiveness lane /rememes fetch hang**: pre-existing dev-server load noise unrelated to this PR (clipboard/i18n only, no data-fetch surface touched); the same class of /rememes slowness was seen on unrelated PRs in the 4.69.0 train.

punk6529:
Integration note: while this PR waited on CI, PR #3099 landed the **desktop** Copy text action on main calling `buildDropClipboardText(drop)` with the pre-PR one-argument signature — a clean textual merge but a type error against this branch's required `locale` parameter (exactly the class of break the up-to-date branch protection exists to catch; an admin merge would have broken main).

Resolved in 59041f3e0 after merging main in: the desktop action now passes its already-resolved `useBrowserLocale()` value to `buildDropClipboardText`, its test asserts the two-argument payload, and the stale comment referencing the old mobile close-on-failure behavior is corrected. `typecheck:ci` and all 5 copy/clipboard suites (28 tests) pass on the merged tree locally.

Follow-up (out of scope here): the desktop action still closes silently on clipboard failure; parity with the mobile failure feedback is flagged as a separate task.

PR #3093, opened 2026-07-05
Align profile CMS builder adapter with the real backend contract

punk6529:
## Issue
- The Profile CMS builder's API adapter (`lib/profile-cms/builder/api.ts`) was written against an assumed backend contract before the real backend landed. The real endpoints and response shapes (generated `ApiProfileCms*` models) differ: the wallet gallery snapshot endpoint moved and returns a completely different shape, and draft save/validate return full package/validation-result models instead of the assumed `{draft_id, package_hash}` stub. Saved drafts also had no way to be listed or loaded back.

## Fix
- Re-point the adapter at the real routes and translate the real snake_case wire shapes into the existing frontend models at the adapter boundary, so the wallet parser, curation UI, and editor state flow stay unchanged. Wire the full draft lifecycle (save / server-validate / get-by-id / list-for-profile) and add a minimal saved-drafts panel so drafts are recoverable. Publish remains hard-blocked.

## Changes
- Wallet gallery snapshot: `POST profile-cms/gallery/snapshots` (assumed) is now `POST profile-cms/wallet-gallery/snapshot` with body `{ wallets: string[] }`. New `lib/profile-cms/builder/gallery-normalize.ts` maps the real response (`generated_at` epoch millis, `source: "indexed_ownership"`, `block_reference`, resolved/unresolved `wallets`, assets with `media`/`flags`, `excluded_assets`, `totals`) into the existing `WalletGallerySnapshot` review model: collections derived by grouping on `collection_key`, media state ready/partial/missing from `media.image`/previews/animations, owner-scoped asset ids (same ERC1155 token can be indexed once per owned wallet), unresolved-wallet and truncation warnings surfaced through review warning codes. Fixture fallback when the builder API flag is disabled is preserved.
- Draft lifecycle: save draft (`POST profile-cms/packages`) and server validate (`POST profile-cms/packages/validate`) now parse the real `ApiProfileCmsPackage` / `ApiProfileCmsValidationResult` responses; `GET profile-cms/packages/:id` and `GET profile-cms/profiles/:profile_id/packages` are wired via `lib/profile-cms/builder/package-normalize.ts` (the on-wire `package` key is read directly since the generated client aliases it to `_package`). Loaded packages are validated against the local V1 schema before entering editor state.
- Builder UI: new "Saved drafts" panel (refresh + per-draft load) in `components/profile-cms-builder/ProfileCmsBuilder.tsx`; reads are gated behind the builder API flag and the connected-profile owner check, load failures and invalid server validation results are surfaced honestly instead of pretending success.
- Publish: the existing client-side hard block is untouched (verified by tests).
- i18n: en-US keys only, for the drafts panel, the invalid-server-validation message, and the two new snapshot warnings. Other locales intentionally fall back to en-US per the existing builder locale policy (separate lane).
- Docs: `ops/workstreams/profile-native-cms-roadmap/builder-mvp-integration-assumptions.md` updated from "expected" to the real contract.

## Validation
- `jest __tests__/lib/profile-cms/builder __tests__/components/profile-cms-builder`: 6 suites, 59 tests, all passing, including new adapter response-mapping tests with realistic backend-shaped snake_case fixtures and a drafts list/load component flow.
- `tsc --noEmit -p tsconfig.typecheck.json`: clean.
- ESLint on changed + new files (`--max-warnings=0`): clean.
- `knip`: no unused exports.
- Not tested: live requests against a deployed backend (the builder API feature flag is off in all environments; the contract was verified against the backend handlers, API service, and generated OpenAPI models on backend main).

## Risk
- Level: Low
- Why: the touched surface is the hidden, feature-flagged builder route; with the flag off the fixture path is unchanged. Publish remains blocked, so no write path to production content changes.
- Rollback: revert this commit; no migrations or config changes.

## Review Notes
- The generated `ApiProfileCmsPackage` client model aliases the wire `package` field to `_package`; the adapter defines a `ProfileCmsPackageWire` type reading the raw `package` key on purpose — see the comment in `package-normalize.ts`.
- `createBuilderStateFromPackage` still imports every loaded draft as the homepage template; loading a wallet-gallery draft re-edits it as homepage blocks. Known MVP limitation, follow-up candidate for the builder lane.
- Snapshot asset ids are owner-scoped (`contract:token:owner`) to keep curation ids unique when the same token is held by multiple connected wallets.

### DIFF
--- a/__tests__/components/profile-cms-builder/ProfileCmsBuilder.test.tsx
+++ b/__tests__/components/profile-cms-builder/ProfileCmsBuilder.test.tsx
@@ -11,13 +11,18 @@ import type { ReactNode } from "react";
 
 import { useAuth } from "@/components/auth/Auth";
 import ProfileCmsBuilder from "@/components/profile-cms-builder/ProfileCmsBuilder";
-import { commonApiPost } from "@/services/api/common-api";
+import {
+  buildCmsPackageCandidate,
+  createDefaultCmsBuilderState,
+} from "@/lib/profile-cms/builder/package";
+import { commonApiFetch, commonApiPost } from "@/services/api/common-api";
 
 jest.mock("@/components/auth/Auth", () => ({
   useAuth: jest.fn(),
 }));
 
 jest.mock("@/services/api/common-api", () => ({
+  commonApiFetch: jest.fn(),
   commonApiPost: jest.fn(),
 }));
 
@@ -46,6 +51,7 @@ jest.mock("next/navigation", () => ({
 
 const useAuthMock = useAuth as jest.Mock;
 const commonApiPostMock = commonApiPost as jest.Mock;
+const commonApiFetchMock = commonApiFetch as jest.Mock;
 const createObjectUrlMock = jest.fn(() => "blob:cms-export");
 const revokeObjectUrlMock = jest.fn();
 const NativeBlob = globalThis.Blob;
@@ -315,10 +321,10 @@ describe("ProfileCmsBuilder", () => {
 
     expect(commonApiPostMock).not.toHaveBeenCalled();
     expect(
-      screen.getByText(
+      screen.getAllByText(
         "Connect as this profile before using backend builder actions."
-      )
-    ).toBeInTheDocument();
+      ).length
+    ).toBeGreaterThan(0);
   });
 
   it("keeps publish honest when backend writes are disabled", async () => {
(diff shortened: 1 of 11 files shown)


punk6529:
Review follow-ups addressed in f6e7983b0: rejected server validation is now modelled as `ok: false` (code `server_validation_invalid`) carrying the blocking issues and the validation target hash/draft id, with the UI special-cases removed; Sonar S7735 negated ternary inverted; resolved-ENS wallet mapping documented and covered by a mixed resolved/unresolved fixture test; saved-draft load-failure branch now has a component test.

**Deferred:** a hash-mismatch indicator when a loaded draft's stored `package_hash` no longer matches the recomputed local hash. Deferred deliberately — the builder recomputes hashes before every save/validate, and the publish/signing flow (where hash provenance actually gates anything) lands in the stacked `codex/cms-publish-signing` branch, which owns that surface.

punk6529:
Superseded by #3934 and companion 6529-Collections/6529seize-backend#1988, now merged and deployed to production. The consolidated implementation and validation are documented in those PRs.

PR #3095, opened 2026-07-05
Exclude locale dictionaries from Sonar copy-paste detection

punk6529:
One-line analysis-scope change: add `i18n/messages/**` to `sonar.cpd.exclusions`.

**Why**: locale dictionaries share identical key scaffolding across languages; adding the same keys to 4 locale files trips the ≤3% new-code duplication gate (PR #3092 measured 99.2%). This mirrors the existing static-content exclusions in the same file, but narrower: CPD only — unused-import and all other code analysis stays active for these files (it recently caught a real dropped-namespace regression there).

**Why a standalone PR**: SonarCloud automatic analysis reads scope settings from the default branch, so the exclusion must land on main before dependent PRs (#3092) can pass their gate.


punk6529:
Closing: verified via the SonarCloud settings API that this project uses automatic analysis with UI-level scope settings (parentOrigin: INSTANCE) — sonar-project.properties is not read, so this change would have no effect. The duplication exclusion for locale dictionaries needs to be set by a project admin in SonarCloud UI: Administration → General Settings → Analysis Scope → Duplications → add `i18n/messages/**`.

PR #3098, opened 2026-07-05
Mirror backend gallery generator and fix draft round-trip

punk6529:
## Stacking note

This PR is stacked on `codex/cms-builder-api-alignment` (PR #3093 / WS-B),
which wired the real wallet-gallery snapshot endpoint and normalization.
Diff here is intentionally scoped to the gap between the reviewed snapshot
and the generated CMS package. Once #3093 merges to `main`, this PR's base
should retarget to `main`.

## Issue

WS-B left two open items in the wallet-gallery builder path:

1. `buildWalletGalleryCmsPackage` (the temporary FE package-generation
   fallback) used an ad-hoc page/block shape instead of matching the
   backend's own deterministic generator contract.
2. Loading a saved wallet-gallery draft always reconstructed "homepage"
   template state via `createBuilderStateFromPackage`, misreading the
   generated home/collection/NFT pages as authored homepage blocks and
   losing all curation state (hidden/featured/order) and the wallet input.

## Investigation

Checked whether the backend's deterministic snapshot generator
(`profile-cms-gallery-package-generator.ts` in `6529seize-backend`) is
reachable from an API route. It is not: it has its own unit test but no
handler or service calls it. `profile-cms.handlers.ts` only exposes
save/validate/publish/list/get/export/rollback/archive on already-built
packages, and `wallet-gallery.handlers.ts` only exposes the raw snapshot
endpoint (assets/collections), not package generation. So this PR takes the
"mirror the generator's semantics into the frontend" path rather than
wiring a live endpoint.

## Fix

- `buildWalletGalleryCmsPackage` now mirrors the backend generator's
  deterministic shape: collections are grouped from visible assets, empty
  collections are dropped, and collections sort featured-first then
  alphabetically by id (matches `sortCollections`/`prepareGallery` in the
  backend module). A dedicated collections-index page is emitted, per
  collection pages use a `lightbox_gallery` block, and NFT detail pages are
  ordered by the same grouped-collection sequence used for collection pages
  (previously they followed raw snapshot order).
- Generated packages are now tagged with `build_manifest.renderer =
  "6529-cms-gallery-builder-mvp"` and carry the full reviewed snapshot plus
  hidden/featured/order curation embedded in a namespaced field on the
  wallet source packet (the V1 schema's source-packet catchall already
  allows arbitrary nested JSON here).
- `createBuilderStateFromPackage` now detects that renderer marker and
  restores `template: "wallet_gallery"` plus the full gallery editor state
  (wallet input, snapshot, hidden/featured/order) instead of falling through
  to the homepage path. A gallery package without the embedded payload
  (e.g. hand-authored) still resolves to the `wallet_gallery` template with
  a fresh gallery state, rather than losing the template entirely.
- The fixture fallback (`createMockWalletGallerySnapshot`, used when the
  builder API flag is disabled) is unchanged, so existing flag-off tests
  keep working.
- Updated `ops/workstreams/profile-native-cms-roadmap/builder-mvp-integration-assumptions.md`
  documenting that the backend generator remains source of truth, why the
  frontend currently mirrors it instead of calling it, the replacement path
  once it's exposed behind an endpoint, and the round-trip contract.

No new UI strings were added (this only touches package-generation logic and
non-UI test files), so no new i18n keys were needed.

## Changes

- `lib/profile-cms/builder/gallery.ts`: grouped/sorted collection pages,
  collections-index page, `lightbox_gallery` block, NFT page ordering,
  embedded round-trip state on the wallet source packet, extracted a shared
  `buildGalleryPageMetadata` helper to avoid repeating the
  locale/search/robots/last_updated shape across four page builders.
- `lib/profile-cms/builder/package.ts`: `createBuilderStateFromPackage` now
  branches on `isWalletGalleryGeneratedPackage` and restores gallery state
  via `restoreWalletGalleryStateFromPackage` instead of always parsing as
  a homepage template.
- `__tests__/lib/profile-cms/builder/gallery.test.ts`: added a
  hand-translated two-collection fixture mirroring the backend generator's
  own "groups collections and applies hide, feature, and reorder controls"
  test case, plus a hidden-collection test and a source-packet round-trip
  test.
- `__tests__/lib/profile-cms/builder/package.test.ts`: added a full
  generate -> JSON round-trip -> reload test proving the gallery template
  and curation restore correctly, and a fallback test for packages missing
  the embedded payload.
- `__tests__/components/profile-cms-builder/ProfileCmsBuilder.test.tsx`:
  added an end-to-end component test that saves a gallery draft, loads it
  back through the drafts panel, and asserts the gallery tab (site title,
  wallet input field) is restored instead of the homepage editor.

## Validation

Run from the worktree with `node_modules/.bin/<tool>` (pnpm direct usage is
blocked in this repo; the `6529` wrapper always `cd`s to the main repo root,
which doesn't work from a worktree checkout):

- `tsc --noEmit -p tsconfig.typecheck.json` — clean (0 errors).
- `eslint .` (repo-wide) — 0 errors, 0 warnings.
- `prettier --check` on all touched files — clean.
- `knip` — 0 unused exports/files after un-exporting three internal-only
  constants that were briefly exported without external callers.
- `jest __tests__/lib/profile-cms __tests__/components/profile-cms-builder`
  — 118/118 passing, including all new tests above.
- Pre-existing `tsc --noEmit -p tsconfig.json` (the non-CI, test-inclusive
  config) has unrelated errors in `gallery-normalize.test.ts` and
  `gallery.test.ts` around generated `ApiProfileCmsWalletGallery*` enum
  literals and TS discriminated-union narrowing through `expect().toBe()`.
  Confirmed via `git stash` that these predate this PR (present on the
  merged `codex/cms-builder-api-alignment` tip before any of these changes).
  Not touched here since the actual CI gate (`tsconfig.typecheck.json`)
  excludes `__tests__/**` and is clean.

## Risk

- Level: Low
- Why: Changes are scoped to the hidden, feature-flagged CMS builder route's
  gallery-generation and package-load paths. No production publish path,
  auth, or non-CMS-builder code is touched. The embedded round-trip field is
  additive JSON on an already-catchall schema field, so existing consumers
  of `CmsPackageV1` that don't know about it are unaffected.
- Rollback: Revert this commit; WS-B's snapshot/normalization work is
  unaffected since it lives in `gallery-normalize.ts`/`api.ts`, which this
  PR does not modify.

## Review Notes

- The frontend generator intentionally does not import the backend module
  (separate deployables); equivalence with the backend's grouping/sorting
  behavior is locked by a hand-translated jest fixture, not a shared
  import — flagged in the integration-assumptions doc as the thing to
  replace once the backend generator is reachable from an endpoint.
- The embedded round-trip payload on the source packet is a frontend-only
  bookkeeping mechanism (not a durable backend contract) and is called out
  as such in the docs so it isn't mistaken for a permanent wire format.

### DIFF
--- a/__tests__/components/profile-cms-builder/ProfileCmsBuilder.test.tsx
+++ b/__tests__/components/profile-cms-builder/ProfileCmsBuilder.test.tsx
@@ -12,6 +12,11 @@ import type { ReactNode } from "react";
 import { useAuth } from "@/components/auth/Auth";
 import ProfileCmsBuilder from "@/components/profile-cms-builder/ProfileCmsBuilder";
 import {
+  buildWalletGalleryCmsPackage,
+  createMockWalletGallerySnapshot,
+  parseWalletGallerySources,
+} from "@/lib/profile-cms/builder/gallery";
+import {
   buildCmsPackageCandidate,
   createDefaultCmsBuilderState,
 } from "@/lib/profile-cms/builder/package";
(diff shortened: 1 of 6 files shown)


punk6529:
Superseded by #3934 and companion 6529-Collections/6529seize-backend#1988, now merged and deployed to production. The consolidated implementation and validation are documented in those PRs.

PR #3107, opened 2026-07-05
Read use-case lock multicall envelopes via .result, honoring status

### LINKED ISSUE #3078: Delegation use-case lock UI reads multicall envelopes instead of .result

punk6529:
## Issue

Fixes #3078. The use-case lock UI in the collection-delegation screen read wagmi `useReadContracts` (multicall) results as booleans. With the default `allowFailure: true`, `data[i]` is an envelope object `{ error?, result?, status }` — always truthy once data loads — so the UI was systematically wrong:

- every use case rendered ` - LOCKED` plus the global `*` marker once statuses loaded
- the lock/unlock branch compared the envelope with `false` (a branch that can never be true for an object), so the "Unlock use case in All Collections" note replaced the Lock/Unlock button for every use case
- the in-flight toast title always said "Unlocking"
- most seriously: the write argument `!useCaseLockStatuses.data?.[i]` negated a truthy envelope, so `setCollectionUsecaseLock` was always submitted with `lock=false` once statuses loaded — the button could never actually lock a use case

This PR intentionally changes visible behavior; the preceding split PR (#3106) preserved the bug byte-for-byte to keep that diff mechanical. Base is `refactor/collection-delegation-split`; it retargets to `main` when #3106 merges.

## Fix

New `readMulticallBoolean(data, index)` helper in `CollectionDelegation.utils.ts` reads `data[index].result` and honors `status`: an entry only counts as locked when `status === "success"` and `result === true`. Failed or missing entries read as unlocked (unknown), never locked. All seven use-case lock sites in `CollectionDelegationLocks.tsx` now go through it (option labels, asterisk, button-vs-note branch, toast title, button icon/label, and the write argument). The collection-level single reads (`useReadContract`) already returned decoded booleans and are unchanged.

## Before / after

Rendered by feeding the Locks view fixed multicall envelopes (fixtures: use case #2 locked for this collection, #4 failed read, #5 locked at the all-collections scope; everything else decoded unlocked) via a local-only harness page on a dev server against the public API. Same fixtures, both commits:

**Before (split PR code, bug preserved)** — every option LOCKED with asterisk, note shown instead of the button for all three selections:

before

**After (this PR)** — labels come from decoded results; the failed read counts as unlocked; the Lock/Unlock button appears with correct label/icon; the All-Collections note appears only for the genuinely globally locked use case:

after

| Surface | Before (envelope truthiness) | After (decoded result) |
| --- | --- | --- |
| Option label, decoded false | `#3 - Airdrops - LOCKED *` | `#3 - Airdrops - UNLOCKED` |
| Option label, decoded true (this collection) | `#2 - Minting / Allowlist - LOCKED *` | `#2 - Minting / Allowlist - LOCKED` |
| Option label, failed multicall entry | `#4 - Voting / Governance - LOCKED *` | `#4 - Voting / Governance - UNLOCKED` |
| Option label, locked globally | `#5 - Avatar Display - LOCKED *` | `#5 - Avatar Display - LOCKED *` (unchanged, now for the right reason) |
| Selected use case, not locked globally | note "Unlock use case in All Collections" (button unreachable) | `Lock Use Case` / `Unlock Use Case` button per decoded state |
| Toast while writing | always "Unlocking Wallet on Use Case ..." | "Locking"/"Unlocking" per decoded state |
| `setCollectionUsecaseLock` args | always `lock=false` once data loaded | `lock=true` when locking, `lock=false` when unlocking |

## Changes

- `components/delegation/CollectionDelegation.utils.ts`: `MulticallEnvelope` type + `readMulticallBoolean` helper (documented failure semantics).
- `components/delegation/collection-delegation/CollectionDelegationLocks.tsx`: the seven envelope sites now use the helper via two named locals (`selectedUseCaseLocked`, `selectedUseCaseLockedGlobally`) and a per-option `lockedGlobally`; the dead `as unknown as boolean` cast and its bug-preservation comment are gone.
- Tests: `readMulticallBoolean` unit tests (success true/false, failure, missing entries, non-boolean results) and a new `CollectionDelegationLocks.test.tsx` suite pinning option labels, asterisks, toast titles, write args, and the globally-locked note branch against envelope fixtures including failure statuses.

## Validation

- Full Jest suite green: 1,969 suites / 11,127 tests (10 new).
- `tsc --noEmit -p tsconfig.typecheck.json` clean; ESLint `--max-warnings=0` clean; Prettier applied.
- react-doctor on the PR diff: 100/100, no issues.
- Local delegation-readonly Playwright pack against the public API: 30/30 passed.
- Debt ratchet unchanged (`any_casts` 4, the removed double cast was not ratchet-visible).
- Before/after harness run documented above (the harness page itself is local-only and not part of the diff).

Not tested: an on-chain `setCollectionUsecaseLock` transaction (needs a funded wallet and changes chain state). The write-arg correction is pinned by unit tests asserting the exact `writeContract` args instead. Residual risk: the lock write path sees only mock traffic before a human verifies it on-chain — flagged in Review Notes.

## Risk

- Level: Medium-High (wallet write path semantics)
- Why: changes what the lock UI shows and, critically, the boolean submitted to `setCollectionUsecaseLock`. The old behavior could only ever unlock; the new behavior locks/unlocks per decoded on-chain state. Blast radius is confined to the Locks section of `/delegation/[collection]`.
- Rollback: single revert restores the (buggy but known) previous behavior; no data or schema impact.

## Review Notes

- Please scrutinize the lock=true/false write-arg change (`args: [collection, useCase, !selectedUseCaseLocked]`) — it is the one site where wrong logic writes chain state. The unit test "locks an unlocked use case ... lock=true args" pins it.
- A separate pre-existing bug is deliberately NOT fixed here: the special use cases (997/998/999) map to `lockUseCaseIndex` 16/17/18 while the multicall arrays put them at 17/18/19, so their post-selection button state/title read the wrong entry. Filed separately (see the issue linked in the PR conversation) to keep this diff single-concern.
- The `status === "success" && result === true` strictness means a temporarily failing multicall entry shows UNLOCKED rather than an error state. That matches the pre-existing UI vocabulary (there is no error state in this select) and fails toward the less alarming label; flagging in case product prefers an explicit unknown state.

### DIFF
--- a/__tests__/components/delegation/CollectionDelegation.utils.test.ts
+++ b/__tests__/components/delegation/CollectionDelegation.utils.test.ts
@@ -3,6 +3,7 @@ import {
   getDelegationsFromData,
   getParams,
   getReadParams,
+  readMulticallBoolean,
 } from "@/components/delegation/CollectionDelegation.utils";
 import {
   CONSOLIDATION_USE_CASE,
@@ -93,5 +94,42 @@ describe("CollectionDelegation utility functions", () => {
     expect(primary?.wallets[0].expiry).toBe("active - non-expiring");
     expect(subDelegation?.wallets[0].expiry).toBe("active - non-expiring");
     expect(consolidation?.wallets[0].expiry).toBe("active - non-expiring");
+  });
+});
+
+describe("readMulticallBoolean", () => {
+  it("reads true only from a successful envelope with result true", () => {
+    const data = [
+      { status: "success" as const, result: true },
+      { status: "success" as const, result: false },
+    ];
+    expect(readMulticallBoolean(data, 0)).toBe(true);
+    expect(readMulticallBoolean(data, 1)).toBe(false);
+  });
+
+  it("treats failed entries as false regardless of result", () => {
+    const data = [
+      { status: "failure" as const, result: undefined },
+      { status: "failure" as const, result: true },
+    ];
+    expect(readMulticallBoolean(data, 0)).toBe(false);
+    expect(readMulticallBoolean(data, 1)).toBe(false);
+  });
+
+  it("treats missing entries and missing data as false", () => {
+    expect(readMulticallBoolean(undefined, 0)).toBe(false);
+    expect(readMulticallBoolean([], 3)).toBe(false);
+    expect(readMulticallBoolean([undefined], 0)).toBe(false);
+  });
+
+  it("does not treat truthy non-boolean results as locked", () => {
+    const data = [
+      { status: "success" as const, result: {} },
+      { status: "success" as const, result: 1 },
+      { status: "success" as const, result: "true" },
+    ];
+    expect(readMulticallBoolean(data, 0)).toBe(false);
+    expect(readMulticallBoolean(data, 1)).toBe(false);
+    expect(readMulticallBoolean(data, 2)).toBe(false);
   });
 });
--- a/__tests__/components/delegation/CollectionDelegationLocks.test.tsx
+++ b/__tests__/components/delegation/CollectionDelegationLocks.test.tsx
@@ -1,0 +1,254 @@
+jest.mock("react", () => {
+  const actual = jest.requireActual<typeof import("react")>("react");
+  return {
+    ...actual,
+    useEffectEvent: (fn: (...args: unknown[]) => unknown) =>
+      actual.useCallback(fn, []),
+  };
+});
+
+import CollectionDelegationComponent from "@/components/delegation/CollectionDelegation";
+import { ALL_USE_CASES } from "@/components/delegation/delegation-constants";
+import { DELEGATION_ALL_ADDRESS } from "@/constants/constants";
+import { fireEvent, render, screen } from "@testing-library/react";
+import { useReadContract, useReadContracts, useWriteContract } from "wagmi";
+
+jest.mock("wagmi", () => ({
+  useReadContract: jest.fn(),
+  useReadContracts: jest.fn(),
+  useWriteContract: jest.fn(),
+  useEnsName: jest.fn().mockReturnValue({ data: undefined }),
+  useWaitForTransactionReceipt: jest.fn().mockReturnValue({
+    isLoading: false,
+    isSuccess: false,
+    isError: false,
+    error: undefined,
+  }),
+  useChainId: jest.fn().mockReturnValue(1),
+}));
+
+jest.mock("@/components/auth/SeizeConnectContext", () => ({
+  useSeizeConnectContext: () => ({ address: "0x0", isConnected: true }),
+}));
+
+const mockUseReadContract = useReadContract as jest.Mock;
+const mockUseReadContracts = useReadContracts as jest.Mock;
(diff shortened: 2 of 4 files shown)


punk6529:
The special-use-case index mismatch mentioned in Review Notes is now filed as #3108 (with the full index-layout analysis). It is deliberately out of scope here: different bug class, same wallet flow, and it was masked until this PR's envelope fix.

punk6529:
Stacked-PR CI note: app-pr-ci (typecheck/lint/tests/build lanes) and the debt-ratchet job trigger only on PRs targeting main, so this PR currently shows the base-independent checks (DCO, secret scan, Snyk, SonarCloud — all green). When #3106 merges, GitHub retargets this PR to main and the full CI fires; I will re-verify all lanes then. Locally the full Jest suite (1,969/11,127), typecheck, lint, react-doctor 100/100, and the 30-test delegation e2e pack are green on this head.

punk6529:
@prxt6529 — review requested (see the note on #3106; formal request blocked by collaborator status). This is the behavior-changing PR of the stack: please scrutinize the setCollectionUsecaseLock write argument change described in Review Notes.

PR #3110, opened 2026-07-05
Catch generic-argument any in the debt ratchet; type the delegation stragglers

punk6529:
## Issue

Repo-health campaign, dead-patterns tail: the debt ratchet's `any_casts` regex counted only `: any` and `as any`, so generic type arguments (`useState<any>()`, `Record<string, any>`, `<any[]>`) were invisible — the blind spot recorded in `ops/workstreams/repo-health-2026-07/any-exceptions.md` after the Thread G typing wave. Five such sites lived in the delegation modules split by #3106.

Third PR of the CollectionDelegation series (#3106 split -> #3107 envelope fix -> this). Base is `fix/delegation-lock-envelopes`; it retargets as the stack merges.

## Fix

1. **Scanner**: `ANY_CASTS_PATTERN` in `scripts/debt-ratchet.cjs` now also matches `any` as a generic type argument — delimited by `<` or `,` before and `,` `>` `[` `)` `|` `&` after. The delimiter requirements were calibrated against the repo so prose ("as powerful as any art", ", any websites"), placeholders (`<any-string>`), and identifiers (`anyLimit`) never match. The pattern is exported and pinned by new jest cases: a 10-match generic fixture (incl. `Map<any, any>` counting twice and whitespace inside the generic) and a 0-match prose/identifier fixture. Documented remaining blind spot: tuple-member `any` (`[any, string]`) — no current instances, and matching brackets textually would trade a non-existent miss for real false positives (markdown links in comments, comparisons).
2. **Delegation stragglers**: the five `useState<any>` sites the split carried verbatim are typed — `RevokeDelegationRequest` / `BatchRevokeDelegationRequest` for the queued revocation writes in `useDelegationRevocation.ts` (the `loading` flip is part of the shape), and `string[]` for the three disclosure-key arrays in `CollectionDelegationSections.tsx`. The delegation area now counts zero.
3. **Baseline** refreshed in the same change: `any_casts` 4 -> 27. The 23 newly visible sites are pre-existing generic-argument usages across 11 non-delegation files (pepe resolve route, clap animation handles, wave traits casts, and 8 single-site files). They are inventoried per-file in the exceptions ledger as owned burn-down backlog — deliberately NOT typed here (each sits in an area this series has not studied; typing them blind in a stack that must stay reviewable would be scope creep). The ratchet blocks new ones from now on.
4. The thread's run-log entry for the whole series rides along.

## Validation

- New + existing debt-ratchet jest suite green (the old "generic arg is not counted" assertion now asserts the opposite, via the exported pattern rather than a duplicated literal).
- `node scripts/debt-ratchet.cjs` passes against the refreshed baseline; `--details any_casts` shows zero delegation entries.
- `tsc --noEmit -p tsconfig.typecheck.json` clean; ESLint `--max-warnings=0` clean; focused delegation + ratchet suites 19/89 green; full Jest green on the stack (1,969 suites / 11,127 tests).

## Risk

- Level: Low
- Why: scanner + typing-only changes; no runtime behavior change (the revocation-params typing is structural on state that already carried exactly these shapes). The baseline rise is definitional (same debt, better instrument), not a regression.
- Rollback: revert; baseline and ledger revert with it.

## Review Notes

- The regex is the load-bearing piece: see the fixture tests for exactly what it does and does not match. If you know a legitimate `any`-adjacent pattern in this repo it would false-positive on, that is the review comment I want.
- `any_casts` 4 -> 27 in the baseline is intentional and documented in the ledger; the alternative (fixing 23 sites across 11 unfamiliar files in this PR) was rejected as unreviewable scope creep.

### DIFF
--- a/__tests__/scripts/debt-ratchet.test.ts
+++ b/__tests__/scripts/debt-ratchet.test.ts
@@ -4,13 +4,13 @@ import os from "node:os";
 import path from "node:path";
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const { countImportStatements, countLines, countMatches } = require(
-  path.join(process.cwd(), "scripts", "debt-ratchet.cjs")
-) as {
-  countImportStatements: (content: string, packages: string[]) => number;
-  countLines: (content: string) => number;
-  countMatches: (content: string, pattern: RegExp) => number;
-};
+const { ANY_CASTS_PATTERN, countImportStatements, countLines, countMatches } =
+  require(path.join(process.cwd(), "scripts", "debt-ratchet.cjs")) as {
+    ANY_CASTS_PATTERN: RegExp;
+    countImportStatements: (content: string, packages: string[]) => number;
+    countLines: (content: string) => number;
+    countMatches: (content: string, pattern: RegExp) => number;
+  };
 
 const SCRIPT_PATH = path.join(process.cwd(), "scripts", "debt-ratchet.cjs");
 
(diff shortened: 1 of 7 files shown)


punk6529:
Stacked-PR CI note: same as #3107 — full app-pr-ci and the Debt ratchet job fire when this retargets to main after the stack below it merges. Locally on this head: debt-ratchet check green against the refreshed baseline, ratchet jest suite green, typecheck/lint green, full Jest suite green (1,969 suites).

punk6529:
@prxt6529 — review requested (see the note on #3106; formal request blocked by collaborator status). Scanner + typing only; the regex fixture tests are the load-bearing review surface.

PR #3112, opened 2026-07-05
Split app/about/press into _sections row-group components

punk6529:
## Issue

`app/about/press/page.tsx` (2,005 lines) is the first split of the 1,500+ series (roadmap: #3111, Tier S pattern-setter). One WP-scrape page holding a ~260-line meta block plus twelve uniform `fusion-builder-row` press articles in a single JSX expression.

## Fix

Behavior-preserving mechanical split. `page.tsx` keeps the meta/head block, page chrome, post wrapper, and `generateMetadata`; the twelve press rows move verbatim into three co-located `_sections/` fragment components rendered in the original order (underscore dir = out of routing, matching the existing `app/[user]/_lib` convention).

| Module | Content | Lines |
| --- | --- | ---: |
| `page.tsx` (shell) | meta block, chrome, composition, metadata export | 367 |
| `_sections/press-rows-1-4.tsx` | fusion rows 1-4 | 601 |
| `_sections/press-rows-5-8.tsx` | fusion rows 5-8 | 602 |
| `_sections/press-rows-9-12.tsx` | fusion rows 9-12 | 558 |

## Changes

- Fragment components emit no DOM of their own; the moved JSX is line-verbatim (plus Prettier re-indent, whitespace-collapse-equivalent).
- Split executed by a fail-closed script (asserts row boundaries, same-indent balance, and the post-content close before touching anything).
- Debt ratchet baseline updated in the same PR: `oversized_files` 89 -> 88, press page healed off the grandfather list; every new file under the 800 cap.

## Validation

- **Hydrated-DOM parity**: the campaign's `wp-scrape-dom-capture` harness captured `/about/press` on a local dev server against the public API before/after; normalized DOM **byte-identical** (141,524 bytes). Determinism calibrated by double-capture on the base commit (identical). A temporary served-marker check proved the server was rendering the split modules during the "after" capture, then was removed and parity re-verified.
- Typecheck (`tsconfig.typecheck.json`) clean; ESLint `--max-warnings=0` clean on all four files; Prettier applied.
- `jest --findRelatedTests` over the changed files (the CI-equivalent): 1 suite / 6 tests green.
- Production build green; `/about/press` statically generated; sitemap intact.
- `react-doctor:diff`: 0 errors; 12 warnings on `page.tsx`, all pre-existing WP-scrape idioms (raw `<a>`, css `<link>`, giant-JSX) that this PR only relocates.

## Risk

- Level: Low
[...]

PR #3111, opened 2026-07-05
Repo health: 1,500+ line split plan (Thread I roadmap)

punk6529:
## Issue

The repo-health campaign requires behavior-preserving splits of production source files over 1,500 lines, but the oversized inventory includes a mix of durable app code and WordPress-migrated legacy content. Reviewers need the roadmap boundary to avoid spending split effort on content artifacts that may be better handled by CMS migration or legacy-content cleanup.

## Fix

Updated the roadmap at `ops/workstreams/repo-health-2026-07/split-1500-plan.md` so the active split campaign distinguishes:

- source-owned static non-WP pages: still eligible for Tier S splits when they exceed the threshold;
- WordPress-scraped / WordPress-migrated legacy pages: recorded in the inventory, but deliberately deferred;
- application logic, UI, auth, composer, and drop-forge files: still split in risk-ascending order.

## Changes

- Marks the nine current WordPress/Yoast/Avada/Fusion-style page routes, including `app/museum/**`, as deferred instead of active Tier S split targets.
- Replaces the museum decision request with a captured WP-migrated legacy-page boundary.
- Renumbers the active split sequence to start with pure logic and continue through OG image, UI, auth, composer, and drop-forge.
- Keeps Tier S available for future source-owned static non-WP pages.
- Updates the expected ratchet impact for the active sequence from 89 -> 73 oversized files.

## Validation

- Docs-only change.
- `git diff --check`

## Risk

- Level: Low
- Why: roadmap-only scope clarification; no runtime code paths touched.
- Rollback: revert the latest roadmap clarification commit.

## Review Notes

- Please focus on the WP/static boundary: WP-migrated legacy pages are deferred, while source-owned static non-WP pages remain eligible for splitting.
- The application-code module maps and risk ordering remain the main implementation roadmap.

PR #3113, opened 2026-07-05
Render the twin museum-district listings from one shared module

punk6529:
## Issue

Second split of the 1,500+ series (roadmap: #3111, Tier S). `app/om/6529-museum-district/page.tsx` (1,830 lines) and `app/museum/page.tsx` (1,800 lines) are the **same WordPress scrape captured twice** — the 6529 Museum District listing, once linking each wing to its oncyber.io space, once to the internal wing pages. Their wing-card content was 88% line-identical, so a plain per-page section split would trip SonarCloud's new-code duplication gate (the same class of failure the press PR hit and fixed).

## Fix

One shared module renders both pages. `components/museum-district/district-wings-data.tsx` holds the 13 wing cards + 2 headings once — each entry carries both pages' hrefs and both pages' WP numeric class suffixes — and `DistrictWingsSections.tsx` reproduces the scrape markup exactly per variant (om renders external `<a target="_blank">` links, museum renders internal `<Link>`s).

| Module | Content | Lines |
| --- | --- | ---: |
| `app/om/6529-museum-district/page.tsx` (shell) | meta block, chrome, rows 1+3, composition | 558 |
| `app/museum/page.tsx` (shell) | meta block, chrome, rows 1+3, composition | 594 |
| `components/museum-district/district-wings-data.tsx` | the listing data (13 cards, 2 headings, both variants' hrefs/suffixes) | 741 |
| `components/museum-district/DistrictWingsSections.tsx` | variant-driven renderer (frame/title/link/image/heading components) | 275 |

## Changes

- Extraction was **fail-closed template matching**: every skeleton line had to match exactly and every cross-page field except hrefs/numeric suffixes had to be byte-equal. The matcher surfaced (and the design absorbed) the three title-link card variants, optional `srcSet`/`sizes`, incrementing nested-column numbers, and the optional `fusion-text-no-margin` class before anything was written.
- The renderer preserves **attribute order** (SSR serializes JSX attribute order; the parity harness would catch any deviation) and the scrape's stray leading/trailing spaces inside class attributes — kept in constants because `prettier-plugin-tailwindcss` strips them from `className` literals (documented in the module; this plugin behavior is worth knowing for every scrape-page PR in this series).
- Page diffs are surgical: 4 insertions, 2,482 deletions; untouched scrape lines stay byte-exact (deliberately not reformatted).
- Ratchet: `oversized_files` 88 -> 86 (both twins healed); every new module under the 800 cap.
- Note vs the #3111 sequence table: this PR covers the om+museum twins only (their shared-module story is one coherent review); capsule-house + cod follow separately with the same treatment (their 20 uniform artwork tiles share a block type).

## Validation

- **Hydrated-DOM parity on both routes**: normalized captures before/after are **byte-identical** (om 137,678 bytes; museum 137,118 bytes); determinism calibrated by double-capture on base. A temporary marker on the shared component rendered exactly 13x on each route, proving every card renders through it.
- Typecheck clean; ESLint `--max-warnings=0` clean; Prettier applied to the new modules (byte-stable).
- `jest --findRelatedTests`: 2 suites / 15 tests green (both page suites).
- `react-doctor:diff` 97/100: 0 errors; warnings are pre-existing scrape idioms this PR only relocates.
- Production build green; both routes statically generated; sitemap intact.

## Risk

- Level: Low
- Why: static scrape twins, no state/hooks/data-fetching; byte-exact DOM proven per route; blast radius is two routes.

## Review Notes

- The interesting surface is `district-wings-data.tsx`: spot-check a couple of entries against the live pages (e.g. the 1 OF 1 ART card's oncyber href on `/om/6529-museum-district` vs the `/museum/imagined-worlds/` link on `/museum`).
- Museum-CMS-migration overlap is flagged on #3111 (decision requested there); if `app/museum/**` gets held, the om half of this module stands alone unchanged.
- Stacked on #3112 (base `refactor/split-press`) purely so the ratchet baseline chain stays linear; retargets to main when it merges. app-pr-ci runs after retarget (repo gotcha for stacked PRs).
- Review requested from the maintainers team. @prxt6529 — review ask: the twin-page data-module pattern is new in this series; capsule/cod will reuse it, so pattern objections are cheapest here.


punk6529:
Adopted the two protective suggestions from the GLM advisory in 0d12f5695:

- **Data-integrity + variant-divergence**: `__tests__/components/museum-district/DistrictWingsSections.test.tsx` now asserts every `href.om` matches `^https://oncyber\.io/`, every `href.museum` matches `^/museum/`, and at render level that the om variant emits only external `target="_blank"` anchors while the museum variant emits zero.
- **Space-sensitive class constants**: the same suite pins the literal `class` attributes (`"fusion-image-element "` with trailing space; `" fusion-imageframe …"` with leading space) on all 13 rendered cards for both variants, so a future reformat that inlines them back into `className` literals fails the suite instead of silently breaking DOM parity.

On the third item (data file type safety): no `as` casts exist in `district-wings-data.tsx` — entries are typed `readonly DistrictEntry[]` where `numbers` and `href` are `Record<DistrictVariant, …>`, so a missing variant key is a compile error, not a silent `undefined`.

punk6529:
Closing as superseded by the merged museum migration in #3209. The museum and OM district routes now use the current migrated-content implementation, so this legacy shared-listing split no longer applies to `main`. The repo-health split roadmap is complete and the current debt-ratchet baseline is zero.

PR #3120, opened 2026-07-05
Persist the fine-pointer latch so lying browsers boot into desktop mode

punk6529:
## Issue
- Reported by @punk6529 on production after 4.71.x: the mobile drop-actions (⋮) button appears on wave load and then disappears, and there is a window where clicking a post does nothing ('no way to reply by clicking a post').
- Root cause: on capability-lying browsers (his Brave reports `any-pointer: fine=false, any-hover=false` while a trackpad is in use), every page load starts in touch mode; the behavioral latch (PR #3099) then flips the UI to desktop on the first cursor glide. The flip is the flicker, and clicks during the transition land in the dead zone. Reproduced exactly on the composer sandbox with a `hasTouch` context: pre-mouse `Open drop actions` buttons: 2 → post-latch: 0; post-latch hover shows Reply/More actions correctly.

## Fix
- Persist mouse evidence in `localStorage` (`6529-fine-pointer`). The helper hydrates the latch at module load — before the first client render — so machines with mouse history boot directly into desktop mode: no touch-mode first paint, no flip, no dead window.
- First-ever visit on a lying browser still transitions once (nothing can prevent that without truthful capability reporting); storage-blocked contexts degrade to the previous session-only latch.
- Genuine touch devices are untouched: they never latch, so nothing is ever stored.

## Validation
- 22 tests green across the touch-first helper suites, including new cases: hydration from a stored flag at module load (desktop from first read, body attribute set, zero events) and persistence after the trusted-mouse-stream latch.
- Sandbox reproduction of the user's symptoms on current main documented above; the fix removes the transitional state for all loads after the first.

## Risk
- Level: Low
- Why: additive persistence for an existing latch; single storage key; failure modes degrade to current behavior.
- Rollback: revert the commit (users' stored flags become inert).

### DIFF
--- a/__tests__/helpers/touch-first-pointer-latch.test.ts
+++ b/__tests__/helpers/touch-first-pointer-latch.test.ts
@@ -70,6 +70,36 @@ describe("behavioral fine-pointer latch", () => {
   afterEach(() => {
     defineMaxTouchPoints(0);
     document.body.removeAttribute("data-fine-pointer");
+    localStorage.removeItem("6529-fine-pointer");
+  });
+
+  it("hydrates the latch from a previous session at module load", () => {
+    localStorage.setItem("6529-fine-pointer", "1");
+
+    const { helpers, restore } = loadHelpersWithSentinel();
+
+    // Desktop from the very first read — no events required, no flicker.
+    expect(helpers.hasFinePointerCapability()).toBe(true);
+    expect(helpers.hasHoverCapability()).toBe(true);
+    expect(helpers.isTouchFirstEnvironment()).toBe(false);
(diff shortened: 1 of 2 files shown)


punk6529:
Rebased on main after merging PR #3115 (@brookr's touch-synthesized mouse guards — pointed out by @punk6529). The two changes are deliberately sequenced: the guards ensure only a genuine cursor glide can latch, and this PR then persists exactly that guarded evidence. A stored `6529-fine-pointer` flag therefore always represents a real mouse — legacy touch stacks (Win 8-era or any OS-level tap→mouse synthesis) can neither latch nor persist. Combined suites: 26 tests green (guards + hydration/persistence).

PR #3127, opened 2026-07-06
Reveal drop hover actions from pointer events, not CSS :hover

punk6529:
## Issue
- Reported by @punk6529 after 4.71.5: on his Surface/Brave, desktop mode is now stable (latch + persistence working — the ⋮ flashed exactly once and the flag stored), but **hovering a message reveals nothing** — no Reply, no ⋯ — so there is still no reply path.
- Root cause: the desktop action bar renders correctly but reveals via `desktop-hover:group-hover:` CSS. On his browser, the same internal input classification that mis-reports `matchMedia` capabilities also **never activates CSS `:hover`**, although mouse pointer events flow normally (selection and the latch both work off them). Verified: the hydrated desktop mode renders the bar (sandbox: Reply/More visible under Playwright, where `:hover` works) and prod CSS contains both reveal arms — the failure is `:hover` activation itself, which no CSS can fix.

## Fix
- The drop row tracks the cursor with `pointerover`/`pointerout` (mouse-type only, containment check on out so child crossings are no-ops) and passes `forceVisible` to `WaveDropActions` and the grouped hover-timestamp. Pointer events are ground truth on these browsers; CSS `:hover` remains the no-JS fallback for everyone else.
- Precedence preserved: open dropdown > link-card suppression > pointer hover > CSS hover.
- `pointerType` uses the file's existing `getPointerType` default (missing → mouse), consistent with the swipe handlers.

## Validation
- 33 tests green across WaveDrop / WaveDropActions / DropMobileMenuHandler, including new cases: pointerover(mouse) forces the bar visible, pointerout clears it, touch pointerover never reveals, suppression outranks the pointer reveal.
- eslint + debt ratchet clean.

## Risk
- Level: Low-Medium
- Why: adds two pointer handlers per drop row and one boolean state; honest desktops behave identically (JS reveal coincides with CSS reveal).

### DIFF
--- a/__tests__/components/waves/drops/WaveDrop.test.tsx
+++ b/__tests__/components/waves/drops/WaveDrop.test.tsx
@@ -796,6 +796,63 @@ describe("WaveDrop", () => {
     );
   });
 
+  it("reveals desktop actions from mouse pointer events without CSS :hover", () => {
+    // Capability-lying browsers never activate :hover although mouse pointer
+    // events flow — the row's pointerenter must force the action bar visible.
+    isMobileMock.mockReturnValue(false);
+    hasTouchInputMock.mockReturnValue(true);
+    isTouchDeviceMock.mockReturnValue(false);
+    setHoverSupport(true);
+    setViewportWidth(1440);
+
+    renderWithEditingDropProvider(
+      <WaveDrop
+        drop={drop}
+        previousDrop={null}
+        nextDrop={null}
+        showWaveInfo={false}
+        activeDrop={null}
+        showReplyAndQuote={true}
+        location={DropLocation.WAVE}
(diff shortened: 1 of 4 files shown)


punk6529:
GLM advisory disposition (stuck `forceVisible`): mostly handled — a cursor leaving the viewport from inside a row fires `pointerout` with `relatedTarget: null`, and the containment guard treats that as an exit and clears the state. The residual (tab-switch while hovering, then returning with the cursor elsewhere) leaves one stale-but-correct-looking bar until the next pointer event crosses that row, which any real mouse movement does; per-row `visibilitychange`/idle timers aren't worth the churn for a self-healing cosmetic. Deferred.

PR #3136, opened 2026-07-06
Clearer, accessible rep-category errors that name the exact rule

brookr:
## Issue
- The Grant Rep error (screenshot in the rep discussion) was one justified paragraph in a cramped desktop modal that lumped every rule into a catch-all: *"…contains invalid characters, is shorter than one character or is longer than 100 characters. Only letters, numbers, spaces, commas, punctuation, parentheses and single quotes are allowed."* The user couldn't tell which rule they broke, it read poorly, and it wasn't announced to assistive tech.

## Fix
- **Names the exact rule, instantly, localized.** A client-side mirror of the server rules (`components/utils/input/rep-category/repCategoryValidation.ts`) returns a specific i18n key the moment a category is chosen — leading dash / over-length (with the real code-point count) / the specific disallowed characters used, each named. No server round-trip for the common cases; the server stays authoritative.
- **Restructured error surface** (`UserPageRepNewRepError`): bold title → the specific reason → the AI-filter explainer as secondary text; left-aligned and responsive instead of `text-justify` in a tiny box.
- **A11y** (`UserPageErrorWrapper`): `role="alert"` so it's announced on appearance; close control takes a localized `aria-label`/`title`.
- **i18n**: new keys under `rep.categories.validation.*` in the source (`en-US`) dictionary; other locales fall back per the existing pattern for this group.

## Changes
- New: `repCategoryValidation.ts` (+ tests). Rewrote `UserPageRepNewRepError`; `role="alert"` + `closeLabel` on `UserPageErrorWrapper`; wired validation into `UserPageRepNewRepSearch.onRepSelect`; 7 new i18n keys.

## Validation
- Jest: validation suite (every rule → message, no unresolved `{placeholders}`) + updated error-component test (title/message/details, reserved-category path). Lint + Playwright-typecheck clean.
- **Verified live** against the running app with dev-auth on a profile's Grant Rep dialog: `art/photography` → `role="alert"`: *"That category name won't work — These characters can't be used: \"/\". Stick to letters, numbers, spaces, dashes and , . ? ! ' ( )"*; `hey-jude` cleared the error and proceeded. Desktop + mobile viewport screenshots captured.

## Risk
- Level: Low — client-side validation + presentation only; server validation unchanged and still authoritative. Worst case on client/server drift is one extra round-trip, never accepting an invalid category.

## Deployment
- Frontend only, but **deploy after backend #1717 is live**: the copy now tells users a dash is allowed (non-leading); shipping ahead of the backend would promise a rule the API still rejects.

## Review Notes
- **e2e scope (open question):** the new validation is client-side and pre-network, covered here by unit + component tests and verified live. The staging E2E packs are read-only/unauthenticated and structurally can't reach an authenticated Grant Rep flow; the fitting home for an interactive test is the authenticated **sandbox** lane (`test:e2e:*-sandbox`), which would need new rep-flow plumbing in `composerSandboxServer`. Deferred as a possible fast-follow rather than expanding the harness here — flagging for reviewer input.

## Summary by CodeRabbit

* **New Features**
  * Added live validation for REP category input, with clearer messages for invalid length, leading dashes, disallowed characters, and hidden characters.
  * Improved error dialogs with localized titles, close labels, and more helpful details when category suggestions are rejected.

* **Bug Fixes**
  * Search and selection now handle REP category limits more consistently, including Unicode character counting.
  * Updated error messaging shown for reserved or invalid categories to match the latest UI text.


brookr:
Round 1 pushed (fc1db9677) — addressing every lane:

- **6529bot general (unreachable length rule)**: the rule check now runs *before* the search-length guard, so `tooLong` (>100) and leading-dash actually surface; the previously-silent <3 drop shows a new `tooShort` message; lengths compared in code points throughout, reconciling the UTF-16/code-point mismatch.
- **6529bot i18n (English baked into {chars})**: `describeChar` no longer emits "line break"/"tab" — invisible characters now get a locale-neutral `U+XXXX` label, which also fixes the "renders as empty quotes" concern for zero-width/format chars. (Tests updated: `\n`→`U+000A`, ZWSP→`U+200B`.)
- **6529bot WCAG 2.4.7**: the close button gains a `focus-visible` outline (was only clearing the default).
- **Knip**: `REP_CATEGORY_MAX_LENGTH` and `RepCategoryViolation` un-exported (module-internal).
- **Sonar duplication**: the flagged 9 lines were my wrapped message entries forming a token-run with the profileCms block; interleaved wrapped/single entries so no two wrapped pairs are adjacent.
- **GLM "comment inside JSX"**: false positive (valid JS comment before the JSX expression; lint+typecheck passed), but moved above `return` to remove all doubt.
- **6529bot responsiveness**: passed (16/16).

Local: 10/10 tests, lint, knip, playwright-typecheck all clean. The single-message precedence (length → leading dash → disallowed) is now commented as intentional per the other nice-to-have.

brookr:
@coderabbitai review

brookr:
Sonar duplication — decisive fix (b867bf48c): reordering couldn't clear it because the quoted-key run structurally matches the locale dictionaries' key blocks regardless of content. Moved the seven rep-validation messages into their own `objectMessages("rep.categories.validation", { … })` group with **unquoted identifier keys** (the existing newVersionToast pattern), which tokenizes as `IDENT: STRING` rather than `"STRING": "STRING"` and can't match those runs. Also addressed the CodeRabbit nitpick: a literal double-quote character is now labeled in single quotes to avoid an ambiguous quote run.

Local: 10/10 tests, lint, knip all clean. Waiting on the Sonar re-run to confirm density drops under 3%.

PR #3139, opened 2026-07-06
Resync glossary.json with the help corpus

punk6529:
## What

Reruns `./bin/6529 run agent-files:sync` on main. Only `public/glossary.json` changes (+4/-2): the `subscriptions.overview` glossary term picks up the Subscriptions Report fact and the `/tools/subscriptions-report` related path that #3133 added to `ops/help/help-index.json`.

## Why

`__tests__/scripts/sync-agent-files.test.ts` fails on current main (000852152): the artifacts regenerated from the help corpus no longer match the committed `public/` files. Root cause: #3132 merged main mid-branch (bringing in #3133's subscriptions-report corpus records) and regenerated `llms.txt` + `help-index.json` but not `glossary.json`.

PR CI didn't catch it because "Installed app checks" selects suites via `jest --findRelatedTests` over changed files, which follows the module graph and can't trace this test's `fs.readFileSync` of the corpus/artifacts — so corpus edits never select the drift test.

## Verification

- Full jest run on clean main at 000852152: `sync-agent-files.test.ts` fails with exactly this glossary diff (plus one unrelated `NotificationsContext.test.tsx` flake that passes on rerun).
- After regeneration: `sync-agent-files.test.ts` passes (12 tests). `help-index:sync` was also rerun and produced no diff — `public/help-index.json` and `public/llms.txt` were already in sync.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated the glossary entry for “The Memes subscriptions” with clearer guidance.
  * Added a note that the About Subscriptions overview links to the Subscriptions Report.
  * Expanded related links to include the Subscriptions Report page.

PR #3143, opened 2026-07-06
CI: run agent-files drift suite on PRs that touch the help corpus or committed artifacts

punk6529:
## Problem

`__tests__/scripts/sync-agent-files.test.ts` verifies that the committed agent artifacts (`public/glossary.json`, `public/llms.txt`, `public/robots.txt`) match what `6529 run agent-files:sync` generates from `ops/help/help-index.json` + `ops/help/llms.txt.template`. But PR CI selects Jest suites with `jest --findRelatedTests`, which follows the **module import graph** — it cannot trace this suite's `fs.readFileSync` of those data files. Two concrete holes:

1. **Corpus edits** (`ops/help/*.json|.template`): the file list feeding `--findRelatedTests` is also restricted to `*.js/jsx/ts/tsx`, so these files never even enter it. A corpus PR that skipped the sync merged green and broke main's full suite (#3132, hotfixed by #3139 on 2026-07-06).
2. **Artifact-only edits** (`public/llms.txt` etc.): worse — these classify as level-1 `presentational-assets`, `install_required` comes out false, and the **entire Installed app checks job is skipped**, so no step-level fix alone could catch a hand-edited or stale artifact.

Because of hole 2, the fix lives in the CI plan (the same place the workflow already computes its changed-file gates), not in a bash diff inside the jest step.

## Fix

Same mechanism as the existing `reviewbot_contract` check + "Verify reviewbot contract" step:

- **`ops/scripts/testing-strategy.cjs`**: new `agent_files_sync` plan check. Fires when the diff touches `ops/help/**` (except `*.md`), the committed artifacts (`public/glossary.json`, `public/llms.txt`, `public/help-index.json`, `public/robots.txt`), the generators (`scripts/sync-agent-files.cjs`, `scripts/sync-help-index.cjs`, `next-sitemap.config.ts` — the robots block lives there), or the suite itself. It also forces `install` on so artifact-only PRs no longer skip the job. Plan changed-file lists come from `git diff --diff-filter=ACMRD`, so **deletions** of an artifact trigger the check too.
- **`.github/workflows/app-pr-ci.yml`**: expose `agent_files_sync_required` and add a "Verify agent files sync" step running `./bin/6529 run test:no-coverage -- __tests__/scripts/sync-agent-files.test.ts` (~36s standalone), mirroring the reviewbot contract step.
- **`__tests__/scripts/sync-agent-files.test.ts`**: `scripts/sync-help-index.cjs` had the same fs-read blindspot with **no drift test at all** (it's a run-on-load script with no exports). The suite now also asserts `public/help-index.json` matches the corpus byte-for-byte (mirroring the script's trailing-newline behavior), so the new CI step covers that artifact too.
- **`__tests__/scripts/testing-strategy.test.ts`**: four new plan cases — corpus change fires the check + install; artifact-only change fires both despite staying risk level 1; generator script change fires it; `ops/help/README.md` stays in the docs fast lane.
- **`ops/help/README.md`**: previously told corpus editors to run only `help-index:sync` — exactly the #3132 trap. Now documents both sync commands and the CI enforcement.

## Verification

- Both suites pass locally: 51/51 (includes the 5 new tests).
- Workflow YAML parses; eslint clean on `testing-strategy.cjs`.
- This PR itself modifies `ops/scripts/testing-strategy.cjs` and the suite, so both the reviewbot-contract step and the new agent-files step should fire on this very PR — check the run for "Verify agent files sync".

## Known limits

- The related-jest step's `*.js/ts/...` pathspec still can't see `.cjs`/data files generally; this PR fixes the agent-files instance via the plan rather than reworking that step.
- `prebuild` regenerates the artifacts at build time, so production output was never stale — the failure mode was main's full suite going red after merge.

## Summary by CodeRabbit

* **New Features**
  * Added CI checks to detect when help-related published files are out of sync with their source content.
  * Expanded the contributor guidance to include regenerating and committing the updated published help artifacts.

* **Bug Fixes**
  * Improved CI planning so relevant help-content changes now trigger the correct validation steps.
  * Added test coverage for the published help index and sync requirements to catch drift earlier.

### DIFF
--- a/.github/workflows/app-pr-ci.yml
+++ b/.github/workflows/app-pr-ci.yml
@@ -31,6 +31,7 @@ jobs:
       playwright_critical_shell_required: ${{ steps.plan_outputs.outputs.playwright_critical_shell_required }}
       dependency_governance_required: ${{ steps.plan_outputs.outputs.dependency_governance_required }}
       reviewbot_contract_required: ${{ steps.plan_outputs.outputs.reviewbot_contract_required }}
+      agent_files_sync_required: ${{ steps.plan_outputs.outputs.agent_files_sync_required }}
 
     steps:
       - name: Checkout code
@@ -226,6 +227,13 @@ jobs:
       - name: Verify reviewbot contract
         if: needs.plan.outputs.reviewbot_contract_required == 'true'
         run: ./bin/6529 run test:no-coverage -- __tests__/scripts/testing-strategy.test.ts
+
+      # jest --findRelatedTests follows the module import graph, so it cannot
+      # see this suite's fs reads of the help corpus and committed artifacts.
+      # The plan requests it by path whenever those files change.
+      - name: Verify agent files sync
+        if: needs.plan.outputs.agent_files_sync_required == 'true'
+        run: ./bin/6529 run test:no-coverage -- __tests__/scripts/sync-agent-files.test.ts
 
       - name: Lint changed source
         if: needs.plan.outputs.lint_changed_required == 'true'
--- a/__tests__/scripts/sync-agent-files.test.ts
+++ b/__tests__/scripts/sync-agent-files.test.ts
@@ -208,6 +208,9 @@ describe("sync-agent-files", () => {
         path.join(repoRoot, "public/help-index.json"),
         "utf8"
       );
+      // scripts/sync-help-index.cjs publishes the corpus as a byte-for-byte
+      // copy, normalizing only a missing trailing newline. If the publish
+      // step ever reserializes instead, update this assertion with it.
       expect(published).toBe(source.endsWith("\n") ? source : `${source}\n`);
     });
   });
--- a/__tests__/scripts/sync-agent-files.test.ts
+++ b/__tests__/scripts/sync-agent-files.test.ts
@@ -198,6 +198,20 @@ describe("sync-agent-files", () => {
     });
   });
 
+  describe("published help-index.json", () => {
+    it("matches the corpus (run `6529 run help-index:sync` after editing it)", () => {
+      const source = fs.readFileSync(
+        path.join(repoRoot, "ops/help/help-index.json"),
+        "utf8"
+      );
+      const published = fs.readFileSync(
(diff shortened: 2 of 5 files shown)


punk6529:
Verified both "Important" findings from the general lane (`ccb7a32a5f5e`) against the repo; neither is a defect. The two nice-to-haves are applied in 210c6994f.

**1. Generator paths (`scripts/` vs `ops/scripts/`) — paths are correct as written.** The generators live at top-level `scripts/`, not `ops/scripts/`:

- `package.json` wires the commands to exactly these files: `"help-index:sync": "... node scripts/sync-help-index.cjs"`, `"agent-files:sync": "... node scripts/sync-agent-files.cjs"`.
- The suite itself imports `require("../../scripts/sync-agent-files.cjs")` from `__tests__/scripts/`, which resolves to `scripts/sync-agent-files.cjs`.

**2. Published help-index byte-compare — `help-index:sync` is a straight copy, not a reserialize.** From `scripts/sync-help-index.cjs`:

```js
const source = fs.readFileSync(sourcePath, "utf8");
fs.writeFileSync(outputPath, source.endsWith("\n") ? source : `${source}\n`);
```

The assertion mirrors that write exactly (trailing-newline normalization only), and it passed on this PR's own CI run. Comparing against generator output instead isn't practical here without restructuring: `sync-help-index.cjs` validates and writes at module load (no `require.main` guard), so importing it from jest would execute the publish. Added a comment at the assertion documenting the coupling so a future switch to reserialization updates it knowingly.

**Applied nice-to-haves (210c6994f):**

- Every path in `AGENT_FILES_SYNC_FILES` (plus `ops/help/llms.txt.template` for the prefix rule) is now pinned by an `it.each` plan test — removing any entry from the set fails `testing-strategy.test.ts`, including `next-sitemap.config.ts` and both generators.
- The broad non-`.md` prefix match on `ops/help/` is intentional and stays: the directory is a generation-input dir by design (corpus + template), so a false-positive ~36s suite run on a hypothetical future auxiliary file is the right trade against silently missing a new generation input.

PR #3150, opened 2026-07-07
Fix open graph batch failure fan-out

simo6529:
## Issue
- Sentry 6529-FRONTEND-1N reports N+1 `/api/open-graph?url=*` GET spans on `/waves/:wave`.
- Client batching existed, but a failed POST batch fell back to parallel single GET requests.

## Fix
- Remove the single GET fallback after POST batch transport failures and timeouts.
- Let failed batches reject the queued per-link promises and clear their cache entries so later retries can use POST batching again.
- Keep per-url `results` and `errors` behavior for successful POST batch responses.

## Changes
- Simplified link-preview batch resolution to rely on `POST /api/open-graph`.
- Updated unit tests for batch failure, timeout, stalled JSON, error bodies, retry after cache eviction, and no GET fan-out.

## Validation
- `SEIZE_SECURE_INSTALL=1 ./bin/6529 run test -- __tests__/services/api/link-preview-api.test.ts`
- `SEIZE_SECURE_INSTALL=1 ./bin/6529 run lint:changed`
- `SEIZE_SECURE_INSTALL=1 ./bin/6529 run typecheck:changed`
- `git diff --check`

## Risk
- Level: Low
- Why: narrow client API and test change. Failed previews already fall back in the UI; this removes the immediate extra GET retry that caused the Sentry N+1 pattern.
- Rollback: revert the commit to restore the prior fallback behavior.

## Review Notes
- Please focus on whether failed batch transport should retry immediately. This intentionally avoids immediate GET fallback to prevent `/api/open-graph?url=*` fan-out.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link preview fetching so failed or stalled batch requests now fail cleanly instead of silently retrying each link one by one.
  * Updated retry behavior so a later request can succeed after an earlier batch failure.
  * Fixed preview error handling to return the correct error response consistently while avoiding extra unnecessary network requests.

### DIFF
--- a/__tests__/services/api/link-preview-api.test.ts
+++ b/__tests__/services/api/link-preview-api.test.ts
@@ -500,93 +500,87 @@ describe("fetchLinkPreview", () => {
     );
   });
 
-  it("falls back to single GET requests when the batch request fails", async () => {
-    const firstPreview: LinkPreviewResponse = {
-      requestUrl: "https://one.example/article",
-      title: "One",
-    };
-    const secondPreview: LinkPreviewResponse = {
-      requestUrl: "https://two.example/article",
-      title: "Two",
-    };
-    fetchMock
-      .mockRejectedValueOnce(new Error("batch unavailable"))
-      .mockResolvedValueOnce(createResponse(firstPreview))
-      .mockResolvedValueOnce(createResponse(secondPreview));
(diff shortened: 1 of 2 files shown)


simo6529:
@coderabbitai review

PR #3154, opened 2026-07-07
Filter Rabby Mobile user-rejected Sentry noise

simo6529:
## Issue
- Sentry issue 6529-FRONTEND-Q was still capturing Rabby Mobile user-rejected wallet requests as object promise rejections.
- The production event had `code: 4001`, `message: Not Allowed`, Rabby Mobile user-agent context, and `userRejectedRequest` in the serialized stack, but the old filter required both Rabby and rejection evidence in the same stack string.

## Fix
- Split Rabby Mobile context detection from `userRejectedRequest` stack detection.
- Keep the filter narrow: it still requires the exact object rejection shape, code/message pair, Rabby Mobile context or stack context, rejection stack evidence, and no app-owned stack or frame evidence.

## Changes
- Added separate Rabby Mobile stack/context constants.
- Updated the Rabby Mobile user-rejected filter to handle the production Android event shape.
- Added regression coverage for the production-shaped Android event.
- Added guard coverage so app-owned serialized stacks and missing Rabby context remain visible.
- Replaced an existing Coinbase WebSocket message regex in a touched file with explicit string matching to satisfy tight lint.

## Validation
- `./bin/6529 run test:no-coverage -- __tests__/utils/sentry-client-filters.test.ts`
- Commit hook tight lint and formatting checks passed.

## Risk
- Level: Low
- Why: scoped to Sentry client-side filtering for a known wallet rejection signature, with app-owned evidence still preserved.
- Rollback: revert this PR to restore the previous filter behavior.

## Review Notes
- Please focus on whether the filter remains narrow enough to avoid hiding app-owned wallet auth failures.

### DIFF
--- a/__tests__/utils/sentry-client-filters.test.ts
+++ b/__tests__/utils/sentry-client-filters.test.ts
@@ -45,8 +45,15 @@ describe("sentry-client-filters", () => {
     "Error: The provider is disconnected from all chains.\n    at o (chrome-extension://acmacodkjbdgmoleebolmdjonilkdbch/background.js:2:7356292)";
   const rabbyMobileUserRejectedStack =
     "Error: Not Allowed\n    at userRejectedRequest (RabbyMobile://native-bundle/background.js:1:1)";
+  const rabbyMobileAndroidUserRejectedStack = [
+    "EthereumProviderError: Not Allowed",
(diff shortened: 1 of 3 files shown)


simo6529:
@coderabbitai review

PR #3178, opened 2026-07-07
Read profile CMS env flags exclusively via publicEnv

punk6529:
## What

`config/profileCmsBuilderEnv.ts` and `config/profileCmsRuntimeEnv.ts` read `process.env` directly. The `no-restricted-syntax` rule in `eslint.config.mjs` bans `process.env` member access for `**/*.{ts,tsx}`. This PR routes every flag read through the schema-validated `publicEnv` (from `@/config/env`), matching the pattern used elsewhere for public flags — no `process.env` fallback.

## Why it was never caught (correction to the original premise)

The violation was assumed dormant because `lint:changed` only scopes to files changed since `main`. The real reason is stronger: **`baseGlobalIgnores` in `eslint.config.mjs` excludes `config/**` from eslint entirely**, so the rule could never fire on these files — not merely "unchanged since the rule landed." Chronology confirms it: rule added 2025-10-28, `config/**` ignore added 2026-01-06, these files landed 2026-06-17 already ignored. A "full eslint pass" over them is vacuously green (`--no-warn-ignored` skips ignored files silently).

I verified the honest way, forcing the full rule set with `--no-ignore`:
- **before:** 7 `no-restricted-syntax` errors across the two files
- **after:** 0 errors

The broader gap (all of `config/**` and `__tests__/**` invisible to eslint) is out of scope here and tracked separately.

## Changes

- **`config/profileCmsBuilderEnv.ts`** — both builder flags now read only `publicEnv.NEXT_PUBLIC_PROFILE_CMS_BUILDER_*` / `publicEnv.PROFILE_CMS_BUILDER_*`; dropped the four `process.env` reads and the now-unused `_name` arg to the boolean helper.
- **`config/profileCmsRuntimeEnv.ts`** — runtime flags read via `publicEnv`; the non-production guard uses the existing `getNodeEnv()` helper instead of `process.env.NODE_ENV`.
- **`config/env.schema.ts`** — added `PROFILE_CMS_RUNTIME_ENABLED` and `PROFILE_CMS_RUNTIME_FIXTURE_PRIMARY` to `publicEnvSchema` (the four builder keys already existed; the two runtime keys did not). `next.config.ts` bakes every schema key from `process.env` into `PUBLIC_RUNTIME`, so the env vars keep working in dev and deploys with no extra plumbing.
- **4 test suites** (`profile-cms-builder-route`, `profile-cms/builder/api`, `ProfileCmsBuilder`, `profile-cms/runtime/fetcher`) — these toggled flags by mutating `process.env` at call time, which only worked because the modules read it lazily; that path no longer exists. Switched to the repo-standard approach: `jest.mock("@/config/env")` returning a mutable copy of the real `publicEnv`, with each set/delete translated one-for-one. Assertions are unchanged.

## Behavioral note

`PROFILE_CMS_RUNTIME_ENABLED` was previously read live from the server process env per request; it is now baked at build time like every other `publicEnv` flag. This is inherent to the publicEnv-only design and consistent with the other CMS flags. Nothing sets it in deployed environments today (dark launch), so there is no runtime change in practice.

## Verification

- `eslint --no-ignore` on both config files: 7 errors → 0
- jest: 6 profile-CMS suites green (route, builder api, ProfileCmsBuilder, runtime fetcher, profile-cms-route, env.base-endpoint) — 64 tests
- `tsc --noEmit -p tsconfig.typecheck.json`: clean
- prettier: clean
- Re-verified after rebasing onto current `main` (which includes the Profile CMS file-splits and the `TENOR_API_KEY`→`GIPHY_API_KEY` schema rename)

cc @prxt

## Summary by CodeRabbit

* **New Features**
  * Added support for new CMS runtime configuration options, including enabling the runtime and selecting a primary fixture in non-production environments.
* **Bug Fixes**
  * Improved how feature flags are read so CMS builder and runtime behavior now follows the app’s shared configuration consistently.
  * Updated test coverage to verify enabled/disabled CMS builder and runtime flows using the same configuration source as the app.

### DIFF
--- a/__tests__/app/profile-cms-builder-route.test.tsx
+++ b/__tests__/app/profile-cms-builder-route.test.tsx
@@ -1,8 +1,14 @@
 import { render, screen } from "@testing-library/react";
 
 import ProfileCmsBuilderPage from "@/app/[user]/cms/builder/page";
+import { publicEnv } from "@/config/env";
 import { getAppCommonHeaders } from "@/helpers/server.app.helpers";
 import { getUserProfile } from "@/helpers/server.helpers";
+
+jest.mock("@/config/env", () => {
+  const actual = jest.requireActual("@/config/env");
+  return { ...actual, publicEnv: { ...actual.publicEnv } };
+});
 
 jest.mock("@/components/profile-cms-builder/ProfileCmsBuilder", () => ({
   __esModule: true,
@@ -41,8 +47,8 @@ const getUserProfileMock = getUserProfile as jest.Mock;
 describe("profile CMS builder route", () => {
   beforeEach(() => {
     jest.clearAllMocks();
-    delete process.env["PROFILE_CMS_BUILDER_ENABLED"];
-    delete process.env["NEXT_PUBLIC_PROFILE_CMS_BUILDER_ENABLED"];
+    delete publicEnv.PROFILE_CMS_BUILDER_ENABLED;
+    delete publicEnv.NEXT_PUBLIC_PROFILE_CMS_BUILDER_ENABLED;
     getAppCommonHeadersMock.mockResolvedValue({});
     getUserProfileMock.mockResolvedValue({
       id: "profile-punk6529",
@@ -59,7 +65,7 @@ describe("profile CMS builder route", () => {
   });
 
   it("renders the builder for a profile handle when enabled", async () => {
-    process.env["PROFILE_CMS_BUILDER_ENABLED"] = "true";
+    publicEnv.PROFILE_CMS_BUILDER_ENABLED = "true";
 
     const page = await ProfileCmsBuilderPage({
       params: Promise.resolve({ user: "punk6529" }),
@@ -76,7 +82,7 @@ describe("profile CMS builder route", () => {
   });
 
   it("returns not found when the profile lookup fails", async () => {
-    process.env["PROFILE_CMS_BUILDER_ENABLED"] = "true";
+    publicEnv.PROFILE_CMS_BUILDER_ENABLED = "true";
     getUserProfileMock.mockRejectedValueOnce(new Error("not found"));
 
     await expect(
@@ -87,7 +93,7 @@ describe("profile CMS builder route", () => {
   });
 
   it("returns not found when the profile lookup has no profile id", async () => {
-    process.env["PROFILE_CMS_BUILDER_ENABLED"] = "true";
+    publicEnv.PROFILE_CMS_BUILDER_ENABLED = "true";
     getUserProfileMock.mockResolvedValueOnce({ handle: "punk6529" });
 
     await expect(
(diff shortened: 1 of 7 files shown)


punk6529:
Thanks — the runtime-toggle observation on `config/profileCmsRuntimeEnv.ts` is correct and worth pinning down. I investigated the intent and deploy model before deciding, and kept the `publicEnv` approach. Disposition below.

**On the build-time-vs-runtime behavior change (Important):**

The old `process.env["PROFILE_CMS_RUNTIME_ENABLED"]` read was the only per-variable runtime-readable path for this flag, but that capability was never wired up and isn't the app's config model:

- **Nothing sets either flag anywhere.** `grep` across `.github/`, env files, and deploy config finds no assignment of `PROFILE_CMS_RUNTIME_ENABLED` or `PROFILE_CMS_RUNTIME_FIXTURE_PRIMARY` — the runtime bridge is dark-launched.
- **Public config is build-time baked by design.** `next.config.ts` bakes every schema key from `process.env` into `PUBLIC_RUNTIME`; at runtime `config/env.ts` / `config/runtimeConfig.ts` read the whole `PUBLIC_RUNTIME` JSON blob (or the baked `.next/PUBLIC_RUNTIME.json` the prod deploy ships) — there is no per-variable runtime injection for public flags. So `publicEnv` semantics here are consistent with the sibling `PROFILE_CMS_BUILDER_*` flags, which already live in `publicEnv` (including their unprefixed variants).
- **`serverEnv` is not a viable alternative home.** `getServerEnv()` returns `null` whenever `SSR_CLIENT_ID`/`SSR_CLIENT_SECRET` fail validation (both required, and empty in local/unauthenticated dev). Routing the CMS flags through it would make them unreadable exactly when SSR creds are absent and would break the dev fixture workflow (`PROFILE_CMS_RUNTIME_FIXTURE_PRIMARY`).
- **Fail-safe semantics preserved.** `!== "false"` default-enabled is unchanged, and the fixture guard still requires `getNodeEnv() !== "production"`, so no dev fixture path can activate in production.

Net: no current or planned runtime kill-switch depends on the old per-var read, and moving these flags onto the build-time model matches how every other public flag in this app is resolved. I've added a doc comment on the helper making the build-time semantics explicit so a future reader doesn't expect a live per-request toggle. If an ops kill-switch is ever wanted, it needs a dedicated design (the current `serverEnv` all-or-nothing SSR coupling doesn't fit).

**On the unprefixed keys in `publicEnvSchema` (Nice-to-have):**

The two new keys are booleans that join `PROFILE_CMS_BUILDER_ENABLED` / `PROFILE_CMS_BUILDER_API_ENABLED`, which are already unprefixed in `publicEnv` — so this follows existing precedent in the same schema. They are feature gates, not secrets (the security lane concurred), so client visibility via `PUBLIC_RUNTIME` is acceptable.

Verified after rebasing onto current `main`: `eslint --no-ignore` on both config files (0 errors, was 7), the 6 profile-CMS jest suites (64 tests), `tsc`, and prettier all green.

PR #3179, opened 2026-07-07
Fix pnpm test script swallowing appended jest test-path patterns

punk6529:
## Issue
- The `test` script ended with the array-type jest option `--coverageReporters=none`. yargs array options keep absorbing the positionals that follow them — even in `--flag=value` form — so an appended test-path pattern (`6529 run test "MyComponent"`) was swallowed into `coverageReporters` instead of selecting suites: jest ran the full ~1,968-suite battery (~4–6 min) and then crashed `require()`ing the pattern as a coverage-reporter module (`Cannot find module 'MyComponent'`).
- This contaminated a large batch of agent test runs during the 2026-07-06 i18n campaign (see the campaign run-log incident audit) and an earlier agent-files resync session.

## Fix
- Reorder the script so the jest invocation ends with a boolean flag: `jest --coverageReporters=none --silent --verbose=false`. Trailing positionals parse as test-path patterns again, and appended flags are honored. No-argument behavior is unchanged.

## Changes
- `package.json`: flag reorder in the `test` script only.
- `scripts/lint-package-json.cjs`: new rule — fail if any script's final shell command is a jest invocation ending in an array-type jest option (option list pinned from the installed jest-cli 30.4.2; dashed aliases normalized). This linter already gates app-pr-ci, dependency-governance, and the prod deploy workflow, so the hazard cannot be reintroduced silently. The package.json path is now overridable via `LINT_PACKAGE_JSON_PATH` so the linter is testable against fixtures.
- `__tests__/scripts/lint-package-json.test.ts`: new suite (8 tests) covering the existing version-pinning rule and the new trailing-flag rule, including a regression pin that the repository package.json stays clean.
- `AGENTS.md` validation matrix: targeted-test guidance now uses the positional form and warns that flags after a `--` separator are misread (pnpm forwards `--` to jest, and jest reads everything after it as a test-path pattern).

## Audit of other jest entry points
- `test:no-coverage` ends with boolean `--coverage=false` — safe.
- `test-json` / `test-json-changed` end in a follow-up command (`pnpm run test-extract-failed-names`), so appended args never reach jest.
- CI workflows pass positional file lists after boolean `--passWithNoTests` / `--findRelatedTests`, and coverage-floor.yml appends nothing after its jest command — safe by construction.

## Validation
Run from a sibling worktree of this branch (jest cannot match tests under dot-directory checkouts):
- `6529 run test "lint-package-json"` → 1 suite selected, 8/8 pass (pre-fix this form ran the full suite, then crashed).
- `6529 run test "lint-package-json" --cacheDirectory=tmp/jest-verify` → 1 suite, flag honored (cache dir created).
- `6529 run test --testPathPatterns="lint-package-json" --cacheDirectory=tmp/jest-verify` → 1 suite (Jest 30 plural flag form keeps working).
- `6529 run test -- "lint-package-json"` → 1 suite; `6529 run test -- --listTests` → `Pattern: --listTests - 0 matches`, confirming the documented `--` misread.
- No-arg selection unchanged: `--listTests` matches 1,972 test files.
- CI parity: `jest --listTests --findRelatedTests <changed files>` → exactly the new suite.
- `node scripts/lint-package-json.cjs` passes the repo file and fails a fixture carrying the old script line; prettier --check clean on the four changed files.
- Typecheck: the only errors are 5 pre-existing ones in `components/waves/CreateDropGifPicker.tsx` caused by `@giphy/react-components` (added to main this morning) being absent from the local install used for verification — unrelated to this diff; CI typechecks against a fresh install.

## Risk
- Level: Low
- Why: no app code touched; jest flag order is semantically irrelevant to jest itself (no-arg selection verified identical); the new lint rule fires only on the exact hazard shape and passes the current repo file.
- Rollback: revert the commit; no data or deploy impact.

## Review Notes
- The array-option list in the linter is extracted from the installed jest-cli 30.4.2 bundle (`type: 'array'` options). `collectCoverageFrom` is string-type in Jest 30 and deliberately absent.
- The i18n integration branch still carries the old script line together with the old linter (a consistent pair), and its package.json edits don't touch the `test` line — its pre-PR rebase onto main picks both changes up without conflict.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -237,7 +237,10 @@ Prefer focused checks first. Escalate based on blast radius.
 - Broader changed-file validation: `6529 run check:changed`.
 - React, Next.js, JSX, TSX, hooks, routing, or UI state: also run
   `6529 run react-doctor:diff` when available.
-- Unit/integration behavior: targeted `6529 run test -- <pattern>`.
+- Unit/integration behavior: targeted `6529 run test <pattern>` (append flags
+  like `--testPathPatterns=<regex>` or `--cacheDirectory=<dir>` directly; do
+  not insert a `--` separator — jest reads everything after `--` as a
+  test-path pattern, so flags placed there are misread).
 - Playwright/user flows: `6529 run test:e2e` or a targeted Playwright run. The
   Playwright config starts `./bin/6529 run dev` on port `3001`.
 - Build-time, generated models, Next config, env/runtime config, proxy, routing,
--- a/__tests__/scripts/lint-package-json.test.ts
+++ b/__tests__/scripts/lint-package-json.test.ts
@@ -3,7 +3,11 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 
-const SCRIPT_PATH = path.join(process.cwd(), "scripts", "lint-package-json.cjs");
+const SCRIPT_PATH = path.join(
+  process.cwd(),
+  "scripts",
+  "lint-package-json.cjs"
+);
 
 const BASE_PACKAGE_JSON = {
   name: "fixture",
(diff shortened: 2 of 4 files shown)


punk6529:
Re the general-lane nice-to-haves: both are intentional. A segment like `jest ... | tee x` is deliberately not flagged — pnpm appends caller args after the final command in the pipeline (`tee`), so jest never sees them there; piping *into* a trailing jest invocation (`... | jest --reporters`) is still flagged since the array option is the segment's final token. And `test-json`/`test-json-changed` stay exempt precisely because their scripts end in a follow-up command — the test at __tests__/scripts/lint-package-json.test.ts pins that. No code change.

PR #3184, opened 2026-07-07
Revert #3170 native device store — intermittent mobile wave crash (React #310), live on prod

punk6529:
## Incident

`/waves?wave=<id>` on a mobile viewport intermittently hard-crashes to the error boundary ("This page couldn't load") with **React error #310 — "Rendered more hooks than during the previous render"** thrown from Next's app Router. **Reproduced against production** (deployed `f4f98b34f`): the minified stacks reference prod's own `web_build/f4f98b34f.../_next` chunks. Staging E2E has been flagging it since the 12:11Z run today — `tests/social/waves-profile-readonly.spec.ts` › "handles legacy wave query links without mutation", 3/3 retries, in both the 12:11Z and 13:13Z runs (web-mobile-chromium).

## Bisect

Local dev-server bisect with a distilled repro (mobile emulation → `/waves?wave=<id>&serialNo=1`, capture pageerrors; the crash is timing-flaky so each sha got multiple runs):

| sha | PR | runs | verdict |
|---|---|---|---|
| `61a3691f1` (09:37 train) | — | 2 clean | good |
| `047aca266` | #3165 mobile-launch-telemetry | 4/4 clean | **good** |
| `cbeb19d3e` | #3172 (touches only `public/glossary.json`) | n/a | can't be it |
| `f5577ca87` | **#3170 native-device-store** | 3/4 crash | **first bad** |
| `e0314f39e` | #3168 | 2/3 crash | bad |
| `a1fc8ae26` | #3162 | crash | bad |
| `f4695b515` | tailwind cleanup | crash | bad |
| `f4f98b34f` (prod) | #3174 | crash (spec + prod run) | bad |
| main tip | — | crash | bad |
| **this revert** | — | **6/6 clean** | **fixed** |

## Why the store breaks

`hooks/nativeDeviceStore.ts`'s subscribe paths notify other subscribers **synchronously during React's commit**: `subscribeToDeviceInfoStore` → `ensureDeviceInfoListeners()` ends with `update()` → `refreshDeviceInfoSnapshot()` → `notifySubscribers(...)`, and `subscribeToCapacitorStore` → `ensureCapacitorListeners()` → `refreshCapacitorSnapshot()` → `emitCapacitorSnapshot(...)` → notify (plus the async `App.getState()` resolution emitting shortly after). Under Next 16's concurrent rendering on the mobile wave path this interleaves with an in-progress render and corrupts the Router's hook accounting — hence the #310 at `Router`'s `useMemo` rather than in an app component.

[...]

PR #3187, opened 2026-07-07
Fix chat-wave legacy-link crash: restore useSearchParams in useActiveWaveManager

punk6529:
## Incident (second regression from today's deploys)

Following a stale legacy link `6529.io/waves?wave=<id>` into a **chat-type wave** (reproduced with Announcements, `da16201a`) crashes to the "This page couldn't load" error boundary on **all viewports**, ~0.9s after load — React #310 "Rendered more hooks than during the previous render", thrown at **Next's app Router's own `useMemo`** (identified by extracting the minified frame `I` from the production chunk: it's the Router — `useActionQueue`/`canonicalUrl` internals). Direct `/waves/<id>` loads (with or without `serialNo`), the waves list, and in-app clicks are all unaffected. Published as a known issue in the 4.76.1 release notes; this PR is the promised fix.

Distinct from the #3170 crash reverted in #3184 — that one hit the spec-covered regular-wave flow; this one only triggers on the server-`redirect()` soft-nav into chat waves, which the staging E2E spec misses because its wave discovery draws from the regular-waves list region.

## Attribution (production-build bisect)

Local production builds + an 4-8-run crash probe per sha (the crash is timing-flaky; single runs lie). Shas that contain #3170 got its revert overlaid (`git revert --no-commit -m 1 f5577ca87`) so the two same-symptom bugs don't confound:

| sha (+overlay where applicable) | contains | result |
|---|---|---|
| `61a3691f1` (pre-window base) | — | clean 4/4 |
| `f4f98b34f` | ...through #3174 | clean 4/4 |
| `663887b91` | + #3171, #3175 | clean 4/4 |
| `8aa1d8453` | + #3143, #3179, #3176, **#3173** | **crash 4/4** |
| prod `499f3977c` | everything | crash 5/5 |

#3143/#3179/#3176 are non-runtime (CI, package.json script, ops tooling) — **#3173 is the only runtime change in the guilty range**.

## Mechanism and fix

#3173 replaced `useSearchParams` in `useActiveWaveManager` with a `useSyncExternalStore` over **monkey-patched `history.pushState`/`replaceState`**. Next's Router calls those methods during its own render/commit work — e.g. reconciling the `redirect()` target from `app/waves/page.tsx`'s legacy-link handler — and the store's snapshot/notification interplay re-enters React mid-render, corrupting the Router's hook bookkeeping.

Two fixes were tested against production builds:

- **Deferring the patched dispatch to a coalesced macrotask: still crashes 4/4.** The re-entrancy isn't only the synchronous event — React's own render-phase store-consistency check over the changing `location.search` snapshot appears sufficient to trigger it.
- **Restoring the `useSearchParams` implementation (this PR): 0/8 crash runs**, plus 6/6 clean on an earlier equivalent build. Restores the exact pre-#3173 file and test (the `hasActiveWaveDropTarget` API from #3168 is retained — it predates #3173). Net −160 lines; everything else in #3173 (wallet startup gate, navigation history, i18n) is untouched.

Unit suite passes (3/3). If the `useSearchParams`→store swap in #3173 was motivated by the Suspense/CSR-bailout lint rule, that need should be re-approached with a Suspense boundary rather than history patching — happy to help design it (cc @simo6529).

## Post-merge plan

Deploy train immediately after merge (staging → E2E → prod), verify the repro is clean against production, then a follow-up release note.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how the app responds to changes in the page URL, making active wave state update more reliably.
  * Preserved existing URL-based wave behavior, including handling of serial numbers, while simplifying navigation tracking.
* **Tests**
  * Updated automated coverage to match the new URL handling behavior and remove outdated checks.

### DIFF
--- a/__tests__/contexts/wave/hooks/useActiveWaveManager.test.tsx
+++ b/__tests__/contexts/wave/hooks/useActiveWaveManager.test.tsx
@@ -1,6 +1,6 @@
 import { renderHook, act, waitFor } from "@testing-library/react";
 import { useActiveWaveManager } from "@/contexts/wave/hooks/useActiveWaveManager";
-import { usePathname } from "next/navigation";
+import { usePathname, useSearchParams } from "next/navigation";
 
 jest.mock("@/hooks/useDeviceInfo", () => ({
   __esModule: true,
@@ -18,22 +18,10 @@ jest.mock("next/navigation", () => ({
     replace: jest.fn(),
   })),
   usePathname: jest.fn(),
+  useSearchParams: jest.fn(),
 }));
 
 describe("useActiveWaveManager", () => {
-  const originalPushState = globalThis.history.pushState;
-  const originalReplaceState = globalThis.history.replaceState;
-
-  afterEach(() => {
-    globalThis.history.pushState = originalPushState;
-    globalThis.history.replaceState = originalReplaceState;
-    delete (
-      globalThis.window as Window & {
-        __locationStatePatch?: unknown;
-      }
-    ).__locationStatePatch;
-  });
-
   beforeEach(() => {
     jest.restoreAllMocks();
     globalThis.history.replaceState(null, "", "http://localhost/");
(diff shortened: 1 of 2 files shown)


punk6529:
Superseded by #3188. The CI build failure here was `useSearchParams() should be wrapped in a suspense boundary` on every static page — #3173 also restructured the provider/Suspense tree, so plainly restoring useSearchParams cannot build. Two further surgical variants (deferred dispatch; popstate-only store) still crashed 8/8 against production builds, so #3188 reverts the whole merge. Full analysis there.

PR #3190, opened 2026-07-07
test(e2e): pack manifest as single source of truth with sync + runner

punk6529:
## Issue

The E2E pack list is hand-copied in five places (package.json `test:e2e*` block, the staging workflow's bash array, prod variants, Device Farm page list, README docs) and has already drifted — production runs fewer packs than staging with no record of why. Agents updating one copy reliably miss the others. PR 2 of the E2E architecture campaign (charter: `ops/workstreams/e2e-architecture-2026-07/`).

## Fix

`tests/packs.manifest.ts` becomes the single registry; everything else is generated or resolved from it at runtime.

## Changes

- **`tests/packs.manifest.ts`** — all 46 packs declared once (safety class, environments, triggers, CLI shape). CommonJS-styled TS so the dependency-free `.cjs` checks can `require()` it via Node type stripping (≥ 22.18)
- **`scripts/sync-e2e-manifest.cjs`** — regenerates the package.json script block (in-place per key; interleaved hand-managed scripts like `test:native-evidence*` keep their position) + `tests/README.md` pack tables; `--check` = drift gate
- **`scripts/e2e-packs.cjs`** — manifest-driven runner for workflows (`--env/--trigger/--pack/--shard/--list`); **hard-fails on empty resolution** (no false-green deploy signal); reproduces the per-pack step-summary UX
- **`staging-e2e.yml`** — calls the runner from the deployed checkout, with the legacy inline loop as fallback until `1a-staging` contains the runner (workflow_run executes main's YAML against the deployed SHA — unbounded skew window otherwise)
- **`debt-ratchet.yml`** — new "e2e manifest drift" step; header covers generated-file consistency
- **`sonar-project.properties`** — CPD exclusion for the manifest's repetitive literals (same precedent as locale dictionaries)
- Campaign workstream dir: README / active-context / run-log

## Validation

- **Byte-exact regeneration proof**: sync reported `package.json already in sync` on first run against the untouched block; a Jest test asserts every rendered script equals the package.json value
- Runner resolves exactly today's **12 staging packs in workflow order**; `--env production --trigger cron` resolves the 10 canary packs; empty resolution exits 1
- 13 Jest tests pass (render proof, in-place replacement, drift exit codes, resolution, summary UX)
- `typecheck:playwright`, focused eslint, `deadcode:knip`, `node scripts/debt-ratchet.cjs` all clean; both workflow YAMLs parse

## Risk

Medium (touches the staging deploy signal). Mitigations: legacy fallback branch keeps pre-runner staging SHAs working; empty-resolution hard-fail prevents silent no-op; the runner path will be rehearsed via `workflow_dispatch` of Staging E2E right after merge, before any real deploy relies on it.

## Review Notes

- The generated block ownership is `/^test:e2e($|:)/` — hand edits there now fail the Debt Ratchet drift step with instructions to edit the manifest and rerun `./bin/6529 run e2e-manifest:sync`
- Follow-up (tracked in the workstream): delete the workflow's legacy fallback once `1a-staging` contains this PR

## Summary by CodeRabbit

* **New Features**
  * Added a manifest-driven way to manage and run E2E packs.
  * Staging E2E runs now use the centralized pack runner when available, with clearer reporting and listing support.
  * Added checks to keep E2E pack definitions, scripts, and generated documentation in sync.

* **Bug Fixes**
  * Prevents silent “green” runs when no E2E packs are resolved.
  * Improves failure reporting with clearer summaries and log excerpts.


punk6529:
Addressed the general-lane findings in 09ec1d184:

1. **Bare `--` forwarding**: `parseArgs` now tolerates a stray `--` (test added). Empirically `./bin/6529 run e2e:packs -- --env staging` forwards cleanly through pnpm (verified locally — resolves the 12 packs), but the gate no longer depends on that wrapper behavior.
2. **Fallback spawn path**: now resolves the wrapper absolutely via `path.join(ROOT, "bin", "6529")` — the relative path with `shell:false` on Linux was indeed broken for direct `node scripts/e2e-packs.cjs` runs.
3. **Spawn-failure vs test-failure**: summary rows now label `(failed to spawn: <error>)` so environment misconfiguration triages differently from real test failures.

On the `require-6529-command` guard question: children of the `npm_execpath` spawn inherit `SEIZE_6529_COMMAND=1` from the wrapper that launched the runner, so the guard passes — documented inline at the spawn site.

PR #3217, opened 2026-07-09
Add perpetual ranking mode for rank waves

brookr:
## Issue
- Rank waves cannot be created as a continuous, never-ending leaderboard. The create flow forces at least one winner announcement (decision point), so every Rank wave either ends at its final announcement or repeats announcement cycles forever. There is no way to express "rank continuously, never announce winners, never end" — needed for use cases like TDH-weighted signaling of trusted network nodes, where standings must keep re-evaluating and nothing should ever finalize.
- The backend already supports this shape: a RANK wave with `decisions_strategy: null` gets `next_decision_time = null` and is permanently skipped by the wave-decision execution loop, and `voting.period.max: null` plus empty `outcomes` are accepted. The gap was entirely frontend.

## Fix
- Add a **ranking mode** choice to the Rank wave Dates step: **Announce Winners** (existing scheduled behavior, default) vs **Perpetual Ranking** (no winner announcements, no end date). Rendered as two bordered selection cards (same `CommonBorderedRadioButton` pattern as the wave-type picker) with each mode's meaning explained inline.
- Perpetual mode submits `decisions_strategy: null`, `voting.period.max: null`, `participation.period.max: null`, and `outcomes: []` — the exact shape the backend treats as a perpetual leaderboard.
- Move the **Show outcomes** toggle from the Overview step to the Outcomes step, where the outcome context lives. For perpetual waves the Outcomes step replaces the award picker with an explainer ("Outcome is leaderboard position") and shows the visibility toggle disabled + locked on, since the live leaderboard is the product.

## Changes
- `types/waves.types.ts`: optional `ongoingRanking` flag on `CreateWaveDatesConfig` (omitted/false = existing scheduled flow; fully backwards-compatible).
- `components/waves/create-wave/dates/CreateWaveDatesRank.tsx`: mode selection cards; perpetual hides the Winners Announcements + rolling end-date sections; the configured schedule is preserved when switching modes (non-destructive toggle); re-selecting the active mode is a no-op.
- `helpers/waves/create-wave.helpers.ts`: perpetual Rank payload sends null decision strategy / null period ends / empty outcomes; all effects type-gated to Rank so a stray flag can never alter other wave types.
- `helpers/waves/create-wave.validation.ts`: decision-time validations skipped for perpetual (still validates start-date ordering); outcomes not required for perpetual Rank only.
- `helpers/waves/wave-rules.helpers.ts`: create-flow rules summary shows "Winner announcements: None (ongoing ranking, no end date)" instead of phantom decision rows.
- `components/waves/create-wave/hooks/useWaveConfig.ts`: seeds the flag; keeps `outcomesVisible` on while perpetual (type-gated).
- `components/waves/create-wave/outcomes/CreateWaveOutcomes.tsx` + `overview/CreateWaveDisplaySettings.tsx` + `CreateWaveStepContent.tsx`: Show-outcomes toggle relocation and perpetual outcomes explainer.
- `components/utils/radio/CommonBorderedRadioButton.tsx`: optional `ariaLabel` prop so the sr-only radio input gets an accessible name (used by the new mode cards; other consumers unchanged).
- Copy fix: the Announce Winners description no longer claims the wave always "ends after the final announcement" (contradicted the open-ended recurring option on the same screen).
- E2E: `tests/social/create-wave-sandbox.spec.ts` extended (perpetual Rank full-submit pathway, stale-schedule validation block, mode switching, Approve isolation guard) and the stale chat baseline repaired (Rules step). `tests/support/composerSandboxServer.cjs` gains a strict allowlist validator for the perpetual Rank body — any submit carrying a decision strategy, an end date, or outcomes for this flow is classified an unsafe mutation and fails the suite.

## Validation
- Jest: create-wave suites (dates/outcomes/overview/hooks/helpers/validation/wave-rules) pass — 190+ tests including 17 new ones covering the mode matrix, payload hardening (stray-flag isolation per wave type), validation gates, and summary rows.
- Playwright sandbox lane (local mock API, mutation-guarded): chat baseline, perpetual Rank end-to-end create (POST body accepted by the strict allowlist), scheduled-mode validation block, and Approve isolation — run on `web-desktop-chromium` and `web-mobile-chromium` projects.
- Full `tsc` typecheck and ESLint (zero warnings) on all changed files.
- Not tested: real-device runs (AWS Device Farm packs are read-only against deployed environments, so they can only exercise this feature after it reaches staging; the mobile-web Playwright project provides the pre-merge device-class evidence).

## Risk
- Level: Low
- Why: purely additive create-flow option, default off; every consumer of `endDate`/decision fields was traced and is null-safe; payload effects are type-gated to Rank; the backend path (null decision strategy) is long-standing supported behavior. No BE changes, no migrations.
- Rollback: revert the PR; no data shape changes. Waves already created in perpetual mode remain valid backend states independent of this UI.

## Review Notes
- The mode switch intentionally preserves the announcement schedule in config while perpetual is selected (payload/validation/rendering all gate on the flag) so switching back is non-destructive — see `CreateWaveDatesRank.tsx` comment.
- `composerSandboxServer.cjs`'s `isExpectedOpenEndedPeriod` replaces a stale validator that required a numeric period max; chat waves have sent `max: null` since the flow dropped the chat Dates step, and the previously rotted (not-in-CI) sandbox spec is repaired in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a **Perpetual Ranking** mode for Rank waves to support an ongoing leaderboard without winner announcements.
  * Updated wave creation UI copy and schedule/rules display to reflect perpetual vs scheduled behavior.
  * Automatically hides outcomes controls where they don’t apply and removes incompatible submission options.
* **Bug Fixes**
  * Perpetual Ranking now correctly clears the end date and aligns saved/derived settings with the selected mode.
  * Validation and wave-creation payloads now match perpetual Rank expectations.
* **Improvements**
  * Enhanced accessibility for bordered radio controls (optional labels/names).


brookr:
Addressed all review findings in d6a7ec753:

**6529bot general (Important) — `outcomesVisible` asymmetry:** agreed, and resolved by removing the state-forcing effect entirely. The stored preference is now never rewritten: the perpetual Outcomes step renders the toggle locked-visible without touching config, and the submit path (`getCreateWaveDisplayMetadataRequests`) skips the hidden-outcomes metadata for perpetual rank waves only. Switching back to Announce Winners restores the user's prior choice — fully symmetric with the schedule preservation. Covered by new hook, component, and metadata-helper tests (including a stray-flag Approve case).

**6529bot general (Important) — `getCreateWaveTimeLockMs` for perpetual:** verified. The helper derives the lock solely from `voting.timeWeighted` (never from the preserved decision schedule), so no stale-schedule leak is possible. A non-null lock with time-weighted voting enabled is *intentional* for perpetual waves: `time_lock_ms` gates the live time-weighted leaderboard rank in the backend (`drop-voting.db.ts` joins leaderboard entries on `time_lock_ms > 0`), not just decision snapshots — so vote smoothing remains meaningful for a wave that never resolves. Added the missing unit assertions: default perpetual → `time_lock_ms: null`; perpetual + time-weighted enabled → the configured interval. The sandbox allowlist's `time_lock_ms === null` remains correct for the default-flow e2e walk.

**6529bot i18n (Important + nice-to-haves):** all new user-facing copy is now message-backed via `t(DEFAULT_LOCALE, ...)` with product-meaning keys (`waves.create.rank.mode.*`, `waves.create.outcomes.*`, `waves.rules.schedule.*`), including the SR-only legend and the `ariaLabel` (derived from the same message as the visible title, so they move together). On the design-for-length note: the mode cards already override the truncation wrapper with `tw-whitespace-normal`, so longer translated titles wrap rather than clip.

**6529bot general + WCAG + CodeRabbit — shared radio `name`:** added an optional `name` prop to `CommonBorderedRadioButton` and grouped the two mode radios under `rank-schedule-mode` for native arrow-key semantics. Other consumers unchanged.

**CodeRabbit nitpick — duplicated perpetual condition:** extracted `isPerpetualRank` in `getCreateNewWaveBody`.

Validation: 235 tests green across the affected suites, full typecheck and zero-warning lint clean. The sandbox e2e strings are byte-identical after the i18n extraction (keys resolve to the same en-US literals asserted by both the unit tests and the spec).

brookr:
/6529bot review all

brookr:
/6529bot review wcag

brookr:
Applied the `getCreateWaveEndDate` type gate in the latest commit. One correction to the finding's premise: a stray `ongoingRanking` on a non-Rank wave could not actually null the end date — the check sits after the Chat/Approve early-returns, so only Rank control flow reaches it (the existing "ignores a stray ongoing flag on approve waves" unit test covers this). Applied anyway for consistency with `isPerpetualRank` so the guarantee is local rather than positional. No behavior change; helpers suites green.

analyticsflowee (approved):

simo6529 (approved):

brookr:
@simo6529 One more approval click needed, sorry — the branch update that pulled in main (your #3243) postdated your approval, and the ruleset requires approval of the most recent push. Nothing else changed: the only new commit is the merge of main itself, all 12 checks are green on the current head, and 6529bot re-reviewed the merge commit with no new findings. Auto-merge is armed, so your approval merges it instantly.

simo6529 (approved):

brookr:
@simo6529 Thanks for the re-approval! One last thing: the main ruleset's update restriction means only bypass actors can actually land the merge — auto-merge armed from my side can't update the protected ref. Could you hit the Merge button (merge commit)? Everything is green and approved.

PR #3241, opened 2026-07-10
Fix remaining wave panel edge cases

ragnep:
## Issue

Follow-up to #3235. Five valid review findings arrived after the original PR checks had turned green, and the About tab still had one divider that was brighter than the rest of the panel.

## Fix

This follow-up applies the review corrections on top of the merged panel redesign and aligns the remaining divider with the established low-contrast treatment.

## Notable changes

- shows a trimmed drop creator address when an activity author has no handle
- renders an unknown drop creator as text instead of a broken profile link
- resolves the activity action label at render time while keeping the panel’s documented `en-US` fallback uniform
- uses stable, trimmed voter profile identifiers instead of empty names or `/undefined` links
- gives signed-out viewers an accurate connect state instead of “No votes yet”
- keeps REP creditor visible and accessible fallback labels consistent
- treats repeated profile images as decorative to avoid duplicate accessible names
- matches the About header divider to the other subtle full-width panel dividers
- adds regression coverage for each corrected edge case and supported-locale fallback behavior

## Validation

- focused sidebar and rating tests: 6 suites / 45 tests passed on the final profile-fallback implementation
- focused i18n fallback tests passed for every supported locale
- strict changed-file lint and type checking passed
- React Doctor: 99/100; only the intentional test `next/image` mock warning remains
- documentation link validation: 277 Markdown files checked
- `git diff --check`

## Risk

Low. The changes are limited to fallback rendering, localized accessible copy, the signed-out voters state, one divider token, and their tests.

## Review notes

`useWaveTopVoters` deliberately disables its query when no connected profile handle is available, so the signed-out connect state reflects the existing data path rather than hiding a successful public fetch. The right panel remains uniformly on the documented `en-US` fallback until locale-specific panel catalogs are enabled.


ragnep:
@coderabbitai review

PR #3252, opened 2026-07-13
Filter Sentry route parameterization cyclic JSON noise

simo6529:
## Issue
- Sentry issue `6529-FRONTEND-CP` continues to receive the known Next.js SDK route-parameterization cyclic JSON error in iOS WKWebView.
- The existing filter matches the error signature, but rejects the event early because Sentry's own parameterization frame can be classified as application-owned.

## Fix
- Exclude only the exact `@sentry/nextjs` route-parameterization frame from app-frame detection.
- Continue reporting generic cyclic JSON failures and any event with another application-owned frame.

## Changes
- Update the route-parameterization predicate to evaluate app ownership after removing exact SDK parameterization frames.
- Add the observed `/notifications` production event shape, including frame metadata and Mobile Safari WKWebView context, as a regression fixture.
- Model the observed line, column, and context-line frame fields in the local Sentry event type.

## Validation
- `./bin/6529 run test:no-coverage --testPathPatterns='(__tests__/utils/sentry-client-filters|__tests__/instrumentation-client)\.test\.ts$' --runInBand` — 2 suites and 262 tests passed.
- `./bin/6529 run lint:changed` — passed.
- `./bin/6529 run lint:changed:tight` — passed.
- `./bin/6529 run typecheck:changed` — passed for the changed TypeScript files.

## Risk
- Level: Low
- Why: The change remains gated by the exact error type, message, mechanism, native stringify frame, Sentry package path, and iOS WKWebView context.
- Rollback: Revert the commit to restore the previous filter behavior.

## Review Notes
- Please focus on the SDK-frame exclusion and the tests that preserve generic cyclic JSON errors and real app-owned frames.
- This filters the known SDK exception before transport; it does not prevent the SDK from throwing it.

### DIFF
--- a/__tests__/utils/sentry-client-filters.test.ts
+++ b/__tests__/utils/sentry-client-filters.test.ts
@@ -822,76 +822,81 @@ describe("sentry-client-filters", () => {
       ...overrides,
     });
 
-  const createObservedSentryCpNotificationsEvent =
-    (): TestSentryClientEvent => ({
-      transaction: "/notifications",
-      exception: {
-        values: [
-          {
-            type: "TypeError",
-            value: __testing.sentryRouteParameterizationMessage,
-            mechanism: {
-              type: __testing.sentryRouteParameterizationMechanismType,
-              handled: false,
-            },
-            stacktrace: {
-              frames: [
-                {
-                  filename:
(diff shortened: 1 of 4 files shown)


simo6529:
@coderabbitai review

PR #3255, opened 2026-07-13
Reduce duplicate notification requests

simo6529:
## Issue
- Sentry issue `6529-FRONTEND-EF` reports an N+1 API call pattern on `/notifications`.
- Marking the active account's notifications as read also invalidated the aggregate query for every other connected account, causing an unnecessary concurrent `v2/notifications?limit=1` fan-out.
- The server prefetched 10 notifications while the client queried 30, so React Query hydration could not reuse the server result.

## Fix
- Keep connected-account unread counts cached when a read mutation only affects the active account.
- Match the server notification prefetch size and query key to the client's 30-item query.

## Changes
- Narrow `invalidateNotifications` to active identity notification queries.
- Preserve the normal polling cadence for other connected accounts.
- Align the authenticated SSR prefetch key and request size with the client infinite query.
- Add regression coverage for both cache invalidation scope and the hydration query key.

## Validation
- `git diff --check`
- Focused Jest execution was attempted, but this isolated worktree has no installed `node_modules` and stopped before Jest with `cross-env: command not found`.
- Pull request CI is the executable validation gate for this branch.

## Risk
- Level: Low
- Why: The change only narrows a cache invalidation that cannot affect non-active JWTs and aligns an existing prefetch with the existing client page size. Cursor pagination, retry behavior, auth guards, proxy guards, and infinite scrolling are unchanged.
- Rollback: Revert the commit to restore the previous invalidation scope and 10-item server prefetch.

## Review Notes
- Verify that active-account unread state still refreshes through `IDENTITY_NOTIFICATIONS`.
- Verify that non-active connected accounts continue updating through their existing 15-second polling query.
- Verify that the server and client notification query keys remain identical for hydration.

### DIFF
--- a/__tests__/components/react-query-wrapper/ReactQueryWrapper.test.tsx
+++ b/__tests__/components/react-query-wrapper/ReactQueryWrapper.test.tsx
@@ -169,13 +169,13 @@ describe("ReactQueryWrapper context", () => {
     });
   });
 
-  it("invalidateNotifications invalidates notification queries", () => {
+  it("invalidateNotifications leaves other connected accounts cached", () => {
     const { client, ctx } = createTestSetup();
     act(() => ctx.invalidateNotifications());
     expect(client.invalidateQueries).toHaveBeenCalledWith({
       queryKey: [QueryKey.IDENTITY_NOTIFICATIONS],
     });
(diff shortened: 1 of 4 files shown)


simo6529:
@coderabbitai review

PR #3282, opened 2026-07-14
Fix useMediaQuery update loop

simo6529:
## Issue
- Sentry issue 6529-FRONTEND-A7 reports `Maximum update depth exceeded` on `/waves/:wave`, with the first-party stack entering `useMediaQuery` from a `MediaQueryList` change event.

## Fix
- Replace effect-driven React state updates with `useSyncExternalStore` over a primitive media-query snapshot.
- Duplicate or synchronous notifications now cause React to re-read the snapshot and only render when the boolean value actually changes.
- Keep a stable `false` server/hydration snapshot.

## Changes
- Preserve modern `addEventListener("change", ...)` subscription cleanup.
- Multiplex the legacy `onchange` fallback so multiple consumers share and safely restore the prior handler.
- Add regression coverage for initial true/false values, server snapshot, changes, duplicate same-value notifications, cleanup, multiple consumers, and legacy behavior.

## Validation
- `6529 run test:no-coverage __tests__/hooks/useMediaQuery.test.tsx --runInBand` (8 tests passed)
- `6529 run test:no-coverage __tests__/components/memes/drops/MemesLeaderboardDrop.test.tsx __tests__/components/waves/leaderboard/gallery/WaveLeaderboardGalleryItem.test.tsx --runInBand` (25 tests passed)
- `6529 run lint:changed`
- `6529 run react-doctor:diff` (100/100, no issues)
- Focused Prettier check and `git diff --check`

## Risk
- Level: Low
- Why: The shared hook affects responsive behavior across several surfaces, but the return type and server snapshot stay unchanged and focused consumer tests pass.
- Rollback: Revert the single implementation commit.

## Review Notes
- Please focus on external-store snapshot stability under repeated synchronous notifications and legacy `onchange` cleanup with multiple consumers.
- No user-facing copy, docs, telemetry, generated files, or dependency metadata changed.

### DIFF
--- a/__tests__/hooks/useMediaQuery.test.tsx
+++ b/__tests__/hooks/useMediaQuery.test.tsx
@@ -1,4 +1,5 @@
 import { act, renderHook } from "@testing-library/react";
+import { hydrateRoot } from "react-dom/client";
 import { renderToString } from "react-dom/server";
 import { useMediaQuery } from "@/hooks/useMediaQuery";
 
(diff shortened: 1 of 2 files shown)


simo6529:
Addressed the hydration concern in `1254297b7` with an explicit `renderToString` → `hydrateRoot` regression test. It confirms the server snapshot renders `false`, the client hydration run creates the browser `MediaQueryList`, subscribes once, and updates to the real `true` snapshot without requiring a query change. The production implementation remains unchanged because the server memo value is not reused by the separate client hydration runtime.

simo6529:
Responsiveness review follow-up: I am deferring the broad-route findings as unrelated to this diff.

- That review ran against superseded head `9384d826eec652344e49f7dad1ce21a2d7f551c9`; the current head is `1254297b79d6c8c26696ed896d266eec4b39270f`.
- `/rememes` and `/meme-calendar` do not import or consume `useMediaQuery`; the report also captured a Next.js `Compiling` overlay on `/rememes`, which is evidence of a dev-server capture race rather than a regression in this hook.
- The changed hook is used by nested Waves item/modal UI, not by the top-level public Waves shell.
- On the current head, the focused desktop route check passed: `6529 run test:e2e tests/critical-shell/critical-shell.spec.ts --project=web-desktop-chromium --workers=1 --grep "keeps the public waves shell mountable"` (1/1).
- The current-head hook suite (including SSR-to-hydration, duplicate notifications, multiple consumers, cleanup, and legacy listeners) and representative Waves/Memes consumer suites are green, as are all latest official PR checks.

No code change is warranted for those stale/unrelated captures.

PR #3284, opened 2026-07-14
Harden Sentry third-party noise filters

simo6529:
## Issue
- Sentry issues 6529-FRONTEND-CP and 6529-FRONTEND-B9 continued receiving production events even though narrow filters were already deployed.
- CP leaked because `beforeSend` sees the raw client event before parsed browser context is available. B9 leaked because the production stack includes a Sentry route-parameterization wrapper frame, so the existing all-`app:///` predicate rejected it.
- 6529-FRONTEND-9N, 6529-FRONTEND-3V, and 6529-FRONTEND-DK are matching injected Twitter, extension, and wallet-relay noise shapes. 6529-FRONTEND-2Y is an intentional sampled first-party network event and must remain reportable.

## Fix
- Match each noise family with the narrowest stable raw-event evidence: exact exception value/type, mechanism where available, and tightly classified frames.
- Keep the existing app-owned source guards and add explicit preservation coverage for first-party cyclic JSON, Twitter, extension messaging, websocket, and network failures.

## Changes
- Accept CP only when the exact Sentry route-parameterization frame and native JSON stringify frame occur in an iOS webview, using the raw User-Agent before enriched browser context exists.
- Allow B9's known Sentry wrapper only alongside Twitter evidence and injected `app:///waves/<uuid>` document frames; reuse that strict family for 9N's exact `currentInset` error. Raw UA matching is anchored to `Twitter/` or `TwitterAndroid/` tokens.
- Add exact 3V filtering for an unhandled `sendMessage` rejection from `app:///injectedScript.bundle.js`, optionally wrapped by the known Sentry browser helper.
- Extend the Coinbase websocket 1006 predicate only for DK's exact `app:///requestRelay.js` / `i.onclose` frame. Repository search found no app-owned symbol.
- Add sanitized raw-event fixtures for CP, B9, 9N, 3V, DK, and 2Y. The 2Y fixture asserts `network_noise_sampled:true` remains unchanged.

## Validation
- `6529 run test --runInBand --runTestsByPath __tests__/utils/sentry-client-filters.test.ts __tests__/instrumentation-client.test.ts` — 282 tests passed.
- `6529 run lint:changed` — passed.
- `6529 run typecheck:changed` — passed.
- `6529 run react-doctor:diff` — 100/100, no issues.

## Risk
- Level: Medium
- Why: The change intentionally drops additional monitoring events, but every new path requires an exact production signature and preservation tests cover nearby first-party failures. No telemetry fields, sampling policy, dependencies, generated files, or PII settings change.
- Rollback: Revert the scoped commits to restore the previous predicates.

## Review Notes
- Please focus on the CP raw User-Agent/webview gate, the strict Twitter wave-document classifier, and the exact Coinbase request-relay frame carve-out.
- No Sentry issues were updated or resolved; they should remain open until the fix is merged and deployed.

### DIFF
--- a/__tests__/fixtures/sentry-noise-filter-hardening.json
+++ b/__tests__/fixtures/sentry-noise-filter-hardening.json
@@ -2,7 +2,10 @@
   "cp": {
     "transaction": "/messages/:wave",
     "request": {
-      "url": "https://6529.io/messages/00000000-0000-4000-8000-000000000001"
+      "url": "https://6529.io/messages/00000000-0000-4000-8000-000000000001",
+      "headers": {
+        "User-Agent": "Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148"
+      }
     },
     "exception": {
       "values": [
--- a/__tests__/fixtures/sentry-noise-filter-hardening.json
+++ b/__tests__/fixtures/sentry-noise-filter-hardening.json
@@ -1,0 +1,201 @@
+{
+  "cp": {
+    "transaction": "/messages/:wave",
+    "request": {
+      "url": "https://6529.io/messages/00000000-0000-4000-8000-000000000001"
+    },
+    "exception": {
+      "values": [
+        {
+          "type": "TypeError",
+          "value": "JSON.stringify cannot serialize cyclic structures.",
+          "mechanism": {
+            "type": "auto.browser.browserapierrors.setTimeout",
+            "handled": false
+          },
+          "stacktrace": {
+            "frames": [
+              {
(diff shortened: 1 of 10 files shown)


simo6529:
@coderabbitai review

simo6529:
Addressed the latest-head general review in `d690e6436`: CP now requires the exact Sentry/native frames plus raw or enriched iOS-WKWebView evidence, with desktop and app-owned-wrapper preservation tests; Twitter raw UA detection is anchored to observed `Twitter/` or `TwitterAndroid/` token boundaries. The exact currentInset/CONFIG values remain intentionally fail-open for any unobserved variation. The Sentry fixture versions preserve the observed raw shapes, while the production matchers remain version-agnostic.

PR #3287, opened 2026-07-14
Harden WalletConnect relay reconnection

simo6529:
## Issue
- Production Sentry issue `6529-FRONTEND-EG` records WalletConnect relay WebSocket failures as unhandled rejections during the `/join-6529` connect flow.
- Event breadcrumbs show the modal opens normally, the user selects WalletConnect, then relay retries accumulate disconnect listeners before a side promise rejects outside AppKit's handled connection promise.

## Fix
- Override only `@walletconnect/universal-provider@2.23.7` to `2.23.9`.
- WalletConnect 2.23.9 contains the upstream relayer fix that serializes and bounds reconnect attempts, prevents unsupervised background connection work, and closes leaked sockets: [WalletConnect PR #7206](https://github.com/WalletConnect/walletconnect-monorepo/pull/7206).
- Keep AppKit's existing WalletConnect error UI and primary connect-promise handling; this does not add a Sentry filter or change app-owned auth/provider behavior.

## Changes
- Add the exact-version pnpm override with an ownership comment.
- Regenerate the lockfile so Reown AppKit 1.8.19 resolves to the aligned WalletConnect 2.23.9 provider/sign-client/core stack.
- Add a regression test that verifies every AppKit 1.8.19 relay edge stays on the patched stack and that none of the aligned 2.23.7 relay packages can re-enter the lockfile.
- Leave the separate WalletConnect 2.21.x connector stack unchanged.

## Validation
- Focused relay regression test: 1 passed.
- Existing AppKit initialization, Wagmi bootstrap, and connect-context suites: 76 passed.
- `./bin/6529 install:frozen`
- `./bin/6529 run lint:diff`
- Scoped Prettier check for the authored test and workspace config.
- Production build passed with the standard public endpoint configuration used by PR CI; generated-file sync left no unrelated diff.
- Dependency graph inspection confirms AppKit resolves the aligned 2.23.9 stack while the separate 2.21.x connector stack remains unchanged.
- Full-repository `format:check` still reports pre-existing formatting warnings outside this change; the authored files pass the scoped check.
- React Doctor was not run because no React, Next.js route, hook, or UI source changed.
- A live relay outage was not forced locally; the failure/retry implementation and its runtime regression coverage are owned by WalletConnect's patched release.

## Risk
- Level: Medium
- Why: this changes a runtime wallet transport stack, but only as a patch-level override of the exact affected 2.23.7 provider line. Direct AppKit/wagmi dependencies and older connector versions are unchanged.
- Rollback: remove the exact override and regenerate the lockfile.

## Review Notes
- Confirm the override remains scoped to `@walletconnect/universal-provider@2.23.7` and does not collapse the separate 2.21.x dependency graph.
- The dependency risk gate correctly requires manual review because five patched transitive package entries replace 2.23.7 without a direct dependency declaration.
- WalletConnect 2.23.9 predates the repository's seven-day minimum release-age window.
- No CSP, RPC, auth, connector configuration, telemetry filter, or user-facing UI change is included.

### DIFF
--- a/__tests__/walletconnect-relay-resilience.test.ts
+++ b/__tests__/walletconnect-relay-resilience.test.ts
@@ -16,8 +16,14 @@ type PnpmWorkspace = {
   overrides?: Record<string, string>;
 };
 
+// Keep the regression boundary explicit so dependency upgrades must re-evaluate
+// whether AppKit still needs this override before changing the expected stack.
 const affectedProvider = "@walletconnect/universal-provider@2.23.7";
 const patchedVersion = "2.23.9";
+const appKitVersion = "1.8.19";
+const appKitSnapshotKey = new RegExp(
+  `^@reown/appkit(?:-[^@]+)?@${appKitVersion.replaceAll(".", "\\.")}(?:\\(|$)`
+);
 
 function readYaml<T>(filename: string): T {
   return parse(readFileSync(resolve(process.cwd(), filename), "utf8")) as T;
@@ -32,9 +38,7 @@ describe("WalletConnect relay resilience", () => {
     expect(lockfile.overrides?.[affectedProvider]).toBe(patchedVersion);
 
     const appKitProviderVersions = Object.entries(lockfile.snapshots ?? {})
-      .filter(
-        ([key]) => key.startsWith("@reown/appkit") && key.includes("@1.8.19")
-      )
+      .filter(([key]) => appKitSnapshotKey.test(key))
       .flatMap(([, snapshot]) => {
         const version =
           snapshot.dependencies?.["@walletconnect/universal-provider"];
--- a/__tests__/walletconnect-relay-resilience.test.ts
+++ b/__tests__/walletconnect-relay-resilience.test.ts
@@ -18,15 +18,24 @@ type PnpmWorkspace = {
 
 // Keep the regression boundary explicit so dependency upgrades must re-evaluate
 // whether AppKit still needs this override before changing the expected stack.
-const affectedProvider = "@walletconnect/universal-provider@2.23.7";
+const affectedVersion = "2.23.7";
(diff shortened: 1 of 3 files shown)


simo6529:
@coderabbitai review

simo6529:
Addressed the snapshot-key precision suggestion in `92248f2e5` by matching only Reown AppKit package keys at the exact `1.8.19` boundary. I kept the AppKit/provider versions explicit intentionally: this is a regression gate, so a future wallet dependency bump should fail until the override is consciously re-evaluated rather than silently accepting a potentially affected stack. The focused test, scoped lint, typecheck, and scoped formatting checks pass on the new head.

simo6529:
Addressed the remaining actionable GLM advisory feedback in `a0bb55385`:

- the regression guard now verifies `core`, `sign-client`, `types`, `universal-provider`, and `utils` are present at 2.23.9 and absent at 2.23.7;
- lockfile paths now resolve from the test location instead of `process.cwd()`;
- `./bin/6529 install:frozen` passes;
- dependency graph inspection confirms AppKit 1.8.19 resolves only the aligned 2.23.9 relay stack while the separate 2.21.x connector graph remains unchanged;
- 2.23.9 was published on 2026-03-27, well outside the seven-day minimum release-age window.

I intentionally kept the exact `@walletconnect/universal-provider@2.23.7` selector instead of `<2.23.9`: the range would also override the existing 2.21.x connectors and broaden this fix beyond the proven AppKit failure boundary. The explicit AppKit/version assertions make a future dependency upgrade fail visibly and require re-evaluation. The repository wrapper pins pnpm 10.33.0, and the frozen install confirms lockfile normalization is stable for that version.

simo6529:
@coderabbitai review

PR #3301, opened 2026-07-15
Wave Create UX: mobile support, safer defaults, clearer feedback

brookr:
## Issue

A Playwright + code audit of the Wave Create flow found that **wave creation was impossible on mobile web** (the waves master-detail layout only showed main content when a wave was selected, so `/waves/create` rendered the wave list instead of the form), plus a cluster of correctness and UX problems: a dangerous default (a rank wave created with defaults ended within ~1 hour), disappearing validation errors, unclear copy, and — surfaced during live iPhone/desktop device testing — keyboard-covered inputs, scroll carry-over between steps, iOS Safari chrome overlapping the footer, focus-zoom, and a rep-outcome category validation mismatch with the backend.

## Fix

Make the create flow fully usable on mobile web and safer by default, add on-device draft autosave/resume, and bring every field's client-side validation in line with the backend so failures surface inline instead of as opaque API errors.

## Changes

**Mobile reachability & layout**
- `WavesMessagesWrapper`: `/waves/create` now qualifies as main content on mobile (the wave list previously always won).
- New `CreateWaveMobileProgress` (step name, "Step N of M", progress bar) shown below `lg` where the vertical step rail is hidden.
- `CreateWaveModal` sizes to `100svh` (small viewport) so the footer/Next button clears iOS Safari's persistent chrome; `globals.css`/`next.config.ts` viewport-height handling.
- Outcome saved-row variants (6 files) reworked grid → flex to stop label/value overlap on narrow screens; Add-to-Timeline and max-submissions label overflow fixed.

**Safer defaults & clearer feedback**
- First winners announcement defaults to **+7 days** (was same-day 23:59, which produced a ~1-hour wave).
- `useWaveConfig`: config changes drop only the errors actually fixed instead of wiping all; new errors still appear only on a forward-step attempt. Errors stay visible and the offending field is scrolled/focused.
- Real Outcomes empty state; Description placeholder "Drop a post" → "Describe your wave".

**On-device drafts (autosave + resume)**
- In-progress waves autosave to `localStorage` (image stripped, total-char cap); Overview shows a collapsible **Saved Drafts** section (starts collapsed, count badge) separated from a **New Wave** card (starts on the wave name). Draft id is ref-based; uses a plain-http-safe random id.

**Mobile input & keyboard ergonomics**
- `inputMode="numeric"` on purely-integer fields, `inputMode="decimal"` on rep/CIC amount fields.
- Keyboard-focus scroll targets the real scroller and re-runs on late keyboard/`visualViewport` resize with an `isConnected` guard.

**Validation alignment (this round)**
- `repCategoryValidation.ts` now mirrors the backend `REP_CATEGORY_PATTERN` exactly: `\p{L}\p{N} ? ! , . ' ( )` + space, 1–100 chars — **dashes removed** (the FE mirror had wrongly allowed dash-except-leading, so a category like `De-Whoop de woo!?` passed the FE check then 400'd on submit). Leading-dash special case and its message key dropped; the duplicate built-in error is suppressed and its typo fixed.

**Mention tagging (this round)** — two independent regressions surfaced by device testing:
- *Data:* Mention search had become wave-scoped (`v2/waves/{waveId}/mention-search`), which returns empty when `waveId` is null — and the Description step composes the wave's opening drop before the wave exists, so no suggestions loaded. `useIdentitiesSearch` now falls back to the global `identities` search for the no-wave case only; every in-wave composer keeps the wave-scoped search unchanged (the create-wave step is the only caller passing a null wave). No backend change.
- *Rendering (mobile):* Even with data, the picker was invisible in the mobile create flow. The typeahead anchor is appended to `document.body`; the mobile flow is `CreateWaveModal`, a `z-[9999]` overlay portalled to body, so the body-level menu (`z-1020`) rendered behind it. Mention z-index is deliberately kept in the 1000–1200 modal band, so the plugin now detects (via `registerRootListener` + an ancestor walk) when its editor sits inside a higher stacking context and lifts only the active menu to `z-[10000]` in that case; all other mention usages are unchanged. Verified on the deployed staging build (detection reports the `z-9999` host in the modal, nothing on the desktop route).

**Scrolling (native app route)**
- Fix the unscrollable create-wave form on the `/waves/create` route (what the native app uses — the mobile-web modal is a separate entry). `WavesLayout` locks document scroll, and only `CreateWaveModal` provided an internal scroll region, so on the route a tall form left Next/Complete off-screen. `CreateWaveFlow` now owns a single shared `overflow-y-auto` scroll region used by both route and modal; `CreateWaveModal` drops its now-redundant scroll; `WavesLayout`'s content container gets `min-h-0` so it can shrink inside the scroll-locked shell. Verified across mobile route, mobile modal, and desktop; `/waves` feed unaffected.

**Consistency**
- `DateAccordion` chevron now matches the app-wide down-collapsed / up-expanded convention.

**i18n**: all new/changed copy behind `waves.create.*`, `wave.create.drafts.*`, `rep.categories.validation.*`.

## Validation

- Sandbox e2e create-wave pack (`tests/social/create-wave-sandbox.spec.ts`, +452 lines): chat, perpetual rank, approve guard, mobile walk (390px: form renders, progress header shows, one step navigable, no horizontal overflow), and multi-draft collapsed-accordion coverage.
- Unit: rep-category validator 7/7; create-wave suite green except the pre-existing timezone-sensitive `CreateWaveDatesApproveEnd` (red only in non-UTC local shells, green in CI, fails identically on clean main).
- `tsc` and ESLint clean on changed files. (Local ESLint infra needed a direct invocation this session because the sandbox has no network to reinstall a mis-detected native resolver binary — a tooling issue, not a code one; CI ESLint is green.)
- Live device testing on iPhone + desktop across the flow; deployed-build staging smoke passed on the prior round.
- Server-side replay confirmed wave create returns 200 for valid payloads; the only failures reproduced were the rep-category dash mismatch (fixed here) and an expired local dev-auth JWT (environmental).

## Risk

- Level: Medium
- Why: the wrapper gate change affects mobile routing for the waves surface (mitigated by the narrow `pathname === "/waves/create"` check + e2e guard); the default-date change and rep-category tightening deliberately change new-wave behavior. Drafts are client-only (`localStorage`), no API/data-shape changes.
- Rollback: revert the merge/commits; no migrations, no API contract changes.

## Review Notes

- Rep-category rules are now identical FE/BE (no dash). If the product ever wants dashes in rep categories, that is a separate deliberate backend `REP_CATEGORY_PATTERN` change (leading dashes read as spreadsheet-formula/negative-sign triggers in rep exports), after which the FE would follow.
- Deliberately out of scope, tracked as follow-ups: Rules→Review step reorder + timezone display, submission partial-failure recovery, and an app-wide chevron-direction standardization pass.

## Summary by CodeRabbit

* **New Features**
  * Added a compact mobile step progress header to the wave creation flow.
  * On-device draft autosave with a collapsible Saved Drafts / New Wave Overview.
* **Improvements**
  * First decision defaults to one week after the minimum timestamp.
  * Numeric/decimal mobile keyboards on number fields; keyboard-aware focus scrolling.
  * Localized wave/outcomes empty states and the create-wave step counter; updated drop description placeholder.
  * Validation errors now persist across non-fixing changes until later validation resolves them; rep-category validation matches the backend.
  * Refreshed various create-wave UI layouts and modal sizing for better responsiveness on iOS Safari.
* **Bug Fixes**
  * Improved `/waves/create` mobile reachability and reduced small-screen overflow/jumping; improved calendar behavior to avoid view resets.
* **Tests**
  * Expanded mobile and flow coverage for empty states, drafts, and error behavior.


brookr:
Addressed the review round in 92b60dabf:

- **Stale `step` in the error re-validation effect** (general, Important): `step` added to the dependency array.
- **Progress bar accessible name** (WCAG, Important): `aria-label` (new `waves.create.progress.label` key) + `aria-valuetext` with the localized step-position text, and the width transition now respects `prefers-reduced-motion`.
- **Route brittleness**: the create-route check tolerates trailing slashes.
- **Outcomes empty-state announcement**: the container announces via `role="alert"` when the outcomes-required error fires (it is the surfaced validation message for that step).
- **en-US spelling**: fulfil → fulfill.
- CI's `Installed app checks` failure was a legacy component test still expecting the old "Drop a post" placeholder — updated.

**Deferred with rationale:**
- *Step label not sourced through i18n* (`CREATE_WAVE_STEPS_LABELS`): this is the same pre-existing constant the desktop step rail renders from, so the mobile header is exactly as translated as the rail it mirrors. Migrating that constant into the message system touches the desktop stepper and its tests, and belongs to the create-flow i18n sweep already queued as a follow-up PR from the audit. Splitting one consumer of the constant into the message system now would leave the two navigators sourcing the same labels from two systems.
- *`t(DEFAULT_LOCALE, ...)` instead of an active-locale accessor*: this matches the repo's current migration pattern across the create flow; switching to a locale hook is part of the same sweep.

PR #3316, opened 2026-07-15
Make Waves and long DMs faster and smoother

ragnep:
## Issue
- Dense Waves feeds mount expensive drops, media, reactions, and sidebar items, so repeated layout work and broad transitions make mobile scrolling costlier than necessary.
- Long direct-message histories also keep too many message subtrees mounted across the shared 5,000px offscreen window, increasing DOM, heap, and frame-time pressure.
- Chat virtualization must preserve exact placeholder height because stale or zero-height measurements can cause visible jumps or hide messages.

## Fix
- Keep the shared Waves behavior and 5,000px render window, while using a tighter 1,200px window only for direct messages.
- Keep each drop's IntersectionObserver stable and track FULL-drop height continuously through one shared ResizeObserver.
- Capture the exact current border-box height immediately before virtualization, so a pending resize notification cannot produce a stale placeholder.
- Ignore non-finite or non-positive heights, cancel unnecessary delayed fallback reads offscreen, and narrow repeated CSS transitions to the properties that actually animate.

## Changes
- Reworked VirtualScrollWrapper measurement, observer lifecycle, cleanup, and render-window configuration.
- Threaded the optional render margin through the existing drop-list component chain and selected the smaller margin only for direct-message waves.
- Added regression coverage across VirtualScrollWrapper, DropsList, and WaveDropsAll for shared observation, stable intersections, margin propagation, zero-height protection, delayed-read cancellation, and final-height preservation.
- Updated repeated Waves action buttons and the collapsed sidebar row from transition-all to transition-colors; removed one no-op wrapper transition.
- No backend, API, generated model, data, routing, dependency, or deployment changes.
- Supersedes the duplicate direct-message work in #3317.

## Validation
- Focused Jest across VirtualScrollWrapper, DropsList, and WaveDropsAll: 91/91 passed.
- Changed-file formatting, lint, and typecheck passed.
- React Doctor: 100/100, no issues across the eight changed React source files.
- Production Next build passed.
- Dense Waves feeds were visually checked at 390×844 and 1440×900, including images/link cards, sidebar states, composer layout, and rapid switching.
- The direct-message change was visually and interactively checked with a synthetic 300-message thread at 412×915 and 1280×720 before consolidation; thread reopen, composer input, and no-selection state worked without update-depth errors.
- Waves 4× CPU-throttled mobile long-scroll profile: p95 frame time 9.2ms, no frame over 50ms, flat DOM/heap, transition-all instances 20 → 1, style-recalculation time -35%, script time -17%, and total task time -48%.
- Direct-message 300-message profile for the preserved 1,200px window:
  - Mobile: visible DOM nodes -35%, retained nodes -40%, heap -4%, p95 frame gap 19.4ms → 17.1ms, typing probe 246ms → 195ms.
  - Desktop: visible DOM nodes -51%, retained nodes -53%, heap -21%, average scroll FPS 23.7 → 30.7, p95 frame gap 91.7ms → 58.4ms, scripting time 9.05s → 7.85s.
  - Desktop typing remained noisy in the opposite direction, so no desktop typing improvement is claimed.
- The aggregate check:changed command reaches the repository-wide Knip stage and reports existing unrelated unused tooling; this PR does not touch those paths.

## Risk
- Level: Medium
- Why: The observer lifecycle is shared, but Waves retain their existing 5,000px behavior and the tighter window is direct-message-only. Dynamic media uses continuous positive-height tracking plus a final exact boundary measurement to preserve scroll geometry.
- Rollback: Revert this PR's five signed commits to restore the previous observer, render-window, and transition behavior.

## Review Notes
- Please focus on reverse-scroll position preservation, dynamic-media placeholder height, shared ResizeObserver cleanup, and the direct-message-only margin boundary.
- A full two-way recycler remains intentionally out of scope because serial jumps, unread anchors, incoming-message pinning, boosted cards, and dynamic media require a broader design.


ragnep:
Addressed the 6529bot lifecycle findings in `940e29a0c`:

- guard singleton clearing so an unsubscribe closure cannot null a different observer instance;
- reset the synchronous measurement fallback whenever FULL height observation restarts;
- add regression coverage for partial unmount and FULL observation restart behavior.

Validation: focused suite 22/22, changed-file lint and typecheck passed, React Doctor 100/100, and `git diff --check` passed.

ragnep:
Also addressed the GLM advisory about legacy ResizeObserver entries in `05a2c19dc`: `borderBoxSize` is now treated as optional at runtime and falls back to `contentRect.height`, with regression coverage. Validation is now 23/23 focused tests plus clean lint and changed-file typecheck.

ragnep:
Consolidated the direct-message performance work from #3317 into this branch.

The integration keeps Waves on the existing 5,000px render window, applies the 1,200px window only to DMs, and preserves the shared ResizeObserver improvement. I also restored an exact final border-box measurement at the virtualization boundary so a pending resize notification cannot create a stale placeholder and visible chat jump.

Current consolidated validation: 90/90 focused tests, changed-file lint/typecheck, React Doctor 100/100, and production build all pass. No backend, generated model, merge, staging, or deployment changes.

ragnep:
Followed up on the latest 6529bot review with explicit re-entry coverage.

- The fallback effect checks `hasMeasuredHeightRef.current` before creating a timer, so a measured FULL drop does not schedule a delayed layout read when it re-enters. The new test proves leave → re-enter → timer advance performs only the one intentional boundary read.
- The existing observation-restart test already exercises FULL → LIGHT → FULL while the drop remains in view and proves the fallback is restored after the ref reset.
- I am intentionally retaining one exact border-box read when a FULL drop leaves the render window. Skipping it based on a recent ResizeObserver notification would reopen the pending-notification/stale-placeholder race and risk a visible chat jump. The read occurs at the 5,000px Waves or 1,200px DM boundary, not every scroll event.

Focused wrapper coverage is now 28/28; the combined focused suite is 91/91.

PR #3329, opened 2026-07-16
Add cyclic JSON timer origin diagnostics

simo6529:
## Issue
- Sentry issue `6529-FRONTEND-CP` reports cyclic `JSON.stringify` failures from `setTimeout`, but the captured stack contains only native `stringify` and Sentry's generic wrapper. The timer callback and scheduling owner are missing.
- The existing WKWebView route-parameterization filter cannot prove third-party ownership and can suppress a genuine application callback with the same stack shape.

## Fix
- Preserve all matching errors instead of applying the unsafe cyclic-JSON suppression. Full retention is intentional while diagnostics are active because keeping only sampled events could hide genuine application failures.
- In production iOS WKWebViews, independently sample 1 in 16 `setTimeout` registrations and retain a bounded, sanitized scheduling stack until that callback runs.
- If the exact cyclic-JSON `TypeError` is thrown synchronously, associate the scheduling evidence with that same error via a `WeakMap`; Sentry remains the only capture path and receives the original error unchanged.

## Changes
- Adds the searchable `cyclic_json_timer_diagnostics=v1` tag and `cyclic_json_timer_schedule_origin` classification.
- Adds machine-readable `cyclicJsonTimerDiagnostics` fields for callback name, WebView family, sample rate, origin classification, and up to eight generated frame coordinates for release/source-map lookup.
- Removes query strings, external origins/extension IDs, timer arguments, wallet/address-shaped identifiers, long token-shaped names, and other dynamic values from the added evidence. Non-HTTP and malformed locations use fixed coarse placeholders.
- Removes the diagnostic wrapper's own frame by stable name rather than a fixed stack position; the production build preserves that name through minification.
- Leaves unsampled timers, string timers, non-WKWebView browsers, unrelated errors, callback arguments, callback receiver, return value, thrown error identity, and timer ID behavior unchanged.

## Validation
- `6529 run test --runInBand --testPathPatterns='(__tests__/utils/cyclicJsonTimerDiagnostics.test.ts|__tests__/instrumentation-client.test.ts)'` — 2 suites, 77 tests passed.
- `6529 run lint:diff` — passed.
- `6529 run lint:uncommitted:tight` — passed.
- `6529 exec tsc --noEmit -p tsconfig.typecheck.json` — passed.
- `6529 run build` and the focused follow-up `6529 run base-build`, with allowed non-secret compile-time endpoint values — passed, including Next.js compile, TypeScript, 265 static pages, and sitemap generation.
- `knip --reporter compact` added no new findings; its nonzero result contains only existing repository-wide unused script/export findings.

## Risk
- Level: Medium
- Why: the change wraps a sampled subset of timer callbacks, but only in production iOS WKWebViews. The wrapper forwards the original arguments and receiver, rethrows the same error, never captures an event itself, and falls back to the unmodified Sentry timer path if diagnostics setup fails.
- Rollback: revert this PR to remove the wrapper and diagnostic enrichment. The underlying error remains visible because this PR deliberately preserves it.

## Review Notes
- Please focus on timer semantic preservation, the independent per-timer sampling trade-off, stack-frame sanitization, and the decision to stop using generic Sentry/WKWebView evidence as a suppression boundary.
- Promise rejections are intentionally outside this diagnostic path: the target cyclic `JSON.stringify` failure is a synchronous callback throw, while rejected promises use a different Sentry mechanism and error object.
- After deployment, query issue `6529-FRONTEND-CP` for `cyclic_json_timer_diagnostics:v1`. Use `extra.cyclicJsonTimerDiagnostics.schedulingFrames` with that event's release to source-map the generated coordinates before deciding between an application fix and a narrow third-party filter.


simo6529:
@coderabbitai review

simo6529:
@coderabbitai review

simo6529:
@coderabbitai review

PR #3332, opened 2026-07-17
Hide the drop edit action once the edit window has passed

brookr:
## Issue
- The edit pencil shows on any of your own drops indefinitely, but the API only accepts edits within its edit window (`MAX_DROP_EDIT_TIME_MS` after the drop was last touched). After that, you can open the editor, compose a change, and lose it to an error on save.

## Fix
- Drops now carry `editable_until` from the API (**paired backend PR: 6529seize-backend#1764** — the exact deadline mirrored from the API's own update guard). The edit action hides when the deadline has passed or editing is disabled, and flips off live at the deadline while mounted. The arrow-up edit-last-message shortcut skips drops it can no longer edit.

## Changes
- `openapi.yaml` + regenerated models: optional nullable `editable_until` on `ApiDrop`/`ApiDropWithoutWave`/`ApiDropV2` (surgical addition — did not sync the rest of the drifted spec).
- `helpers/waves/drop-editability.helpers.ts`: pure `isDropEditableAt` — `undefined` ⇒ legacy API, keep showing (deploy-order safe); `null` ⇒ editing disabled; number ⇒ editable strictly before.
- `hooks/drops/useIsDropEditable.ts`: re-evaluates exactly when the deadline passes.
- `WaveDropActionsEdit`: gated (author check unchanged).
- `getLatestEditableChatDrop`: arrow-up target respects the window.

## Validation
- 10 new unit tests (helper semantics incl. boundary, component gating incl. live flip-off at deadline, non-author). Existing editable-chat-drop suites still pass (19 total). tsc, ESLint, Prettier, debt ratchet clean.

## Risk
- Level: Low
- Why: purely subtractive UI gating; the API remains the enforcer. The `undefined` fallback means this deploys safely before or after the backend change.
- Rollback: revert.

## Review Notes
- Deploy order preference: backend #1764 first, then this — but the legacy fallback makes either order safe.
- The gate mirrors the backend's semantics including 'window unset ⇒ editing disabled ⇒ null'.


brookr:
Both prior findings addressed in 4360e70d9:

- **32-bit setTimeout overflow** (`useIsDropEditable`): the timer now clamps each wait to `2^31 - 1` ms, re-evaluates on fire instead of assuming expiry, and re-arms until the real deadline — a far-future `editable_until` no longer hides a still-valid edit action. New test drives a 30-day deadline past the cap (action stays visible) and across the actual expiry (action hides).
- **Unconditional mount-time `setIsEditable`**: state writes now bail when the value is unchanged (functional update compares previous).

PR #3352, opened 2026-07-17
Filter expected AWS RUM abort failures

simo6529:
## Issue

AWS RUM currently counts proven intentional request cancellations as HTTP failures. That inflates network-failure views and makes real request failures harder to see.

## Fix

Wrap the official AWS RUM Fetch and XHR plugins at their record boundary and drop only these exact HTTP-event cancellation forms:

- error type `AbortError`
- error message `signal is aborted without reason`
- error message `Fetch is aborted`

All successful requests, HTTP status failures, timeouts, generic network errors, near matches, unverified XHR aborts, and non-HTTP telemetry remain recorded.

## Notable changes

- Add a fail-open, non-blocking expected-abort plugin wrapper.
- Keep the existing privacy sanitizer first in the AWS RUM plugin list.
- Route retained Fetch/XHR events through the existing privacy sanitizer.
- Preserve the official Fetch/XHR plugin IDs and lifecycle methods.
- Add focused classifier, delegation, ordering, privacy, and near-match tests.
- Document the exact filter contract in the monitoring docs and telemetry registry.

## Validation

- Focused Jest: 3 suites, 28 tests passed.
- `lint:diff`: passed.
- `lint:uncommitted:tight`: passed.
- `typecheck:ci`: passed.
- React Doctor: 100/100.
- Production Next.js base build: passed with the required public endpoint variables supplied.
- `check:changed`: scoped typecheck and lint stages passed; the repository-wide Knip stage reports pre-existing unused files/exports unrelated to this diff.

## Risk and rollback

Risk is limited to AWS RUM HTTP-event classification; application requests and user behavior are unchanged. The matcher is intentionally exact and fails open. Rollback is a revert of this PR.

## Review notes

Please focus on the exact-match boundary, official-plugin delegation, and plugin ordering. In particular, unverified `XMLHttpRequest abort`, 4xx/5xx responses, timeouts, generic failures, and all non-HTTP events are deliberately retained.

## Summary by CodeRabbit

* **New Features**
  * Updated AWS RUM HTTP telemetry handling by filtering out expected XHR/fetch abort/cancellation events while allowing normal outcomes (including non-2xx responses and timeouts).
  * Ensured monitoring remains non-blocking if classification fails or event recording throws.

* **Bug Fixes**
  * Improved HTTP URL exclusion logic to align with the active XHR/fetch plugin configuration.

* **Tests**
  * Added/expanded coverage for expected-abort dropping behavior, fail-open handling, and non-impact on delegate lifecycle.

* **Documentation**
  * Clarified AWS RUM HTTP filtering rules, privacy boundary behavior, and provider notes.

### DIFF
--- a/__tests__/components/monitoring/AwsRumProvider.test.tsx
+++ b/__tests__/components/monitoring/AwsRumProvider.test.tsx
@@ -2,21 +2,24 @@ import React from "react";
 import { render, screen, waitFor } from "@testing-library/react";
 import { publicEnv } from "@/config/env";
 import AwsRumProvider from "@/components/monitoring/AwsRumProvider";
-import { AwsRum } from "aws-rum-web";
+import { AwsRum, FetchPlugin, XhrPlugin } from "aws-rum-web";
 
 const WAVE_ID = `${"a".repeat(8)}-${"b".repeat(4)}-4${"c".repeat(3)}-a${"d".repeat(3)}-${"e".repeat(12)}`;
 const OTHER_WAVE_ID = `${"f".repeat(8)}-${"1".repeat(4)}-4${"2".repeat(3)}-a${"3".repeat(3)}-${"4".repeat(12)}`;
+const TEST_AUTHOR_ID = `${"1".repeat(8)}-${"2".repeat(4)}-4${"3".repeat(3)}-8${"4".repeat(3)}-${"5".repeat(12)}`;
+const MEDIA_URL = `https://media.example.cloudfront.net/renditions/drops/author_${TEST_AUTHOR_ID}/file.m3u8`;
 let mockPathname = `/waves/${WAVE_ID}`;
+
+type MockPluginContext = {
+  record: (eventType: string, eventData: object) => void;
+  recordPageView: (pageId: string) => void;
+};
 
 jest.mock("next/navigation", () => ({
   usePathname: () => mockPathname,
 }));
 
 jest.mock("aws-rum-web", () => {
-  type MockPluginContext = {
-    record: (eventType: string, eventData: object) => void;
-    recordPageView: (pageId: string) => void;
-  };
   type MockPluginConfig = {
     readonly eventPluginsToLoad?: Array<{
       load?: (context: MockPluginContext) => void;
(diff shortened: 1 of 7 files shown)


simo6529:
@coderabbitai review

simo6529:
Addressed the advisory review on latest head `8e736c4e62`:

- Documented the intentional fail-open/silent downstream-recorder policy in `ops/telemetry/README.md`.
- Kept the exact `AbortError` type match: this is one of the verified cancellation shapes in scope, and the pinned AWS RUM SDK creates HTTP response events and HTTP error events on separate paths, so 4xx/5xx response events are not suppressed by this classifier.
- No extra ordering test was added because the provider test already uses the real `AwsRumExpectedAbortPlugin` wrapper and proves that a retained raw author URL reaches the existing privacy sanitizer while an exact abort does not reach the recorder. The classifier itself does not inspect URLs.

The follow-up is documentation-only; the focused 28-test monitoring suite and tight lint passed on the preceding code head.

simo6529:
Responsiveness review triage: no code change is warranted for this telemetry-only PR.

The reported blocking failures are dev-server on-demand compilation/timeouts on `/meme-lab` and `/rememes`, not behavior touched by this diff. The review's own evidence shows both routes render successfully in other viewport/runtime contexts, and it identifies compile/render timing as the likely cause. This PR changes only AWS RUM event filtering, its focused tests, and monitoring documentation; it does not change layouts, routes, rendering, Capacitor shims, or responsive styles. The production Next.js build also completed successfully locally.

The native shim warnings are likewise browser-simulation limitations reported across every route, not a regression introduced here. I am therefore leaving the scoped implementation unchanged and classifying these findings as baseline/environment failures rather than implementation regressions.

simo6529:
GLM advisory follow-up on latest head c788bdf520:

- Added provider-level coverage for both Fetch and XHR. Each path now proves a retained failure reaches the existing privacy sanitizer without the raw author identifier, while the exact AbortError is dropped.
- The provider mocks are intentionally minimal wiring doubles. Wrapper lifecycle delegation is already covered directly in awsRumExpectedAbort.test.ts, so duplicating that at provider level would not add a distinct regression boundary.
- The pinned aws-rum-web Plugin type defines record(data) with one argument; the two-argument suggestion does not match the installed SDK.
- The classification fail-open Proxy throws specifically when the classifier reads eventData.error, so it exercises the intended failure point.
- Constructor configuration forwarding is already guarded by the existing Fetch/XHR urlsToExclude equality assertion, which reads the actual constructor calls.

Local follow-up validation: focused monitoring Jest 29/29, tight changed-file lint passed, typecheck:ci passed, and git diff --check passed.

PR #3359, opened 2026-07-17
Show direct message reactions as they happen

ragnep:
## Issue

- Reactions added or removed by another participant could remain invisible in an open direct-message thread until the conversation refetched.
- Repeated WebSocket close/reconnect cycles also created delivery gaps for other live updates.

## Fix

- Let the WebSocket provider exclusively own unexpected-close recovery and exponential backoff.
- Keep message listeners registered across connection and authentication state changes while delivering each listener's latest callback.
- Preserve the existing canonical reaction refetch and optimistic reconciliation behavior.

## Changes

- Removed socket status from the initial health-check effect so it no longer cancels a pending provider retry.
- Kept realtime subscribers attached through reconnects and component renders.
- Added focused coverage for listeners registered while disconnected and retained across socket replacement.
- Updated the realtime and reaction documentation.

## Validation

- `6529 run lint:changed`
- `6529 run build`
- `6529 run react-doctor:diff` (100/100 on the source diff)
- Focused WebSocket test suites: 3 passed
- Documentation link validation
- Desktop and mobile local browser QA on an open direct-message thread: reaction controls and counts rendered, with no browser or WebSocket console errors
- A real conversation was not mutated to manufacture a second participant's reaction; deployed staging validation will cover recovery and live behavior where practical

## Risk

- Level: Medium.
- Reason: the code change is small, but it is in the shared WebSocket lifecycle used by all live events.
- Rollback: reverting the commits restores the previous status-driven reconnect and subscription behavior.

## Review Notes

- Please focus on the ownership split between the health hook and provider retry loop, plus subscriber lifetime across reconnects.
- The project uses React 19.2.4; its installed development and production runtimes both export `useEffectEvent` directly.
- The provider subscriber map is independent of socket instances; focused coverage now verifies delivery after a socket is replaced.
- Canonical reaction fetching and the optimistic mutation guard are intentionally unchanged.

## Summary by CodeRabbit

* **Documentation**
  * Clarified authenticated live-update and listener readiness during reconnects.
  * Documented that Waves and Messages resume supported live events immediately after reconnect+authentication.
  * Clarified that events only arrive while the WebSocket is connected, and missed messages aren’t guaranteed to replay.
  * Added guidance that reactions and ratings update live while a wave/DM thread is open.
* **Bug Fixes**
  * Improved real-time subscription behavior so listeners persist across unexpected reconnects.
  * Refined WebSocket health-check and message subscription handling.
* **Tests**
  * Updated and added coverage for reconnect persistence and subscription/unsubscription behavior.

### DIFF
--- a/__tests__/services/useWebSocketMessage.test.ts
+++ b/__tests__/services/useWebSocketMessage.test.ts
@@ -12,27 +12,28 @@ describe('useWebSocketMessage', () => {
   beforeEach(() => {
     subscribe.mockReset();
     unsubscribe.mockReset();
+    subscribe.mockReturnValue(unsubscribe);
     (useWebSocket as jest.Mock).mockReturnValue({
-      subscribe: jest.fn(() => unsubscribe),
+      subscribe,
       status: WebSocketStatus.CONNECTED,
     });
   });
 
-  it('subscribes and unsubscribes when connected', () => {
+  it('subscribes and unsubscribes while mounted', () => {
     const cb = jest.fn();
     const { unmount } = renderHook(() => useWebSocketMessage('TYPE' as any, cb));
-    expect(useWebSocket().subscribe).toHaveBeenCalledWith('TYPE', cb);
+    expect(subscribe).toHaveBeenCalledWith('TYPE', expect.any(Function));
     unmount();
     expect(unsubscribe).toHaveBeenCalled();
   });
 
-  it('does not subscribe when not connected', () => {
+  it('keeps the listener registered while disconnected', () => {
     (useWebSocket as jest.Mock).mockReturnValue({
       subscribe,
       status: WebSocketStatus.DISCONNECTED,
     });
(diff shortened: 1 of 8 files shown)


ragnep:
6529bot follow-up evidence for the latest head:

- `package.json` pins React to `19.2.4`, not React 18.
- React 19.2.4's installed development runtime, production runtime, and type declarations all export `useEffectEvent` under that exact stable name.
- `6529 run typecheck:changed` passes.
- A full `6529 run build` passes under Next.js 16.2.6, including repository-wide lint, production compilation, TypeScript, prerendering, and sitemap generation. This directly exercises the import-resolution path requested in the review.
- The separate subscriber-lifetime concern is covered by the new disconnected → socket replacement → message delivery test, which passes.

The remaining “Needs changes” import concern is based on a React 18 premise that does not apply to this repository's pinned React 19.2.4 runtime. No fallback import is needed.

PR #3414, opened 2026-07-22
Fix Release Bus preflight profile transport

GelatoGenesis:
## Summary
- derive the protected target build profile in its own credential-bearing job
- carry the verified profile to authorization through a same-run immutable artifact
- keep downstream gate/build outputs non-secret so GitHub preserves them across the job boundary
- add focused contract coverage proving the authorization job never receives the HMAC secret

## Validation
- `./bin/6529 run test:no-coverage --maxWorkers=2 --runTestsByPath __tests__/scripts/release-bus-frontend-gate.test.ts`
- `./bin/6529 run typecheck`
- Prettier check on both changed files

Workflow/tooling-only change; no frontend application deployment is required.

### DIFF
--- a/.github/workflows/release-bus-preflight.yml
+++ b/.github/workflows/release-bus-preflight.yml
@@ -93,7 +93,7 @@ jobs:
           name: release-bus-preflight-build-profile-${{ github.run_id }}
           path: ${{ runner.temp }}/release-bus-build-profile.json
           if-no-files-found: error
-          retention-days: 14
+          retention-days: 1
 
   authorize:
     needs: build_profile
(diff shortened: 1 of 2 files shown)


GelatoGenesis:
Addressed the actionable review in `788852636`:

- The trust boundary is the current workflow run, not an unauthenticated cross-run artifact. `download-artifact` has no `run-id` or `repository`, `authorize` has a hard `needs: build_profile` edge, the v4 artifact name is immutable and keyed to `github.run_id`, and `build_profile` is the only job that can run before the consumer and upload that name. A compromised producer would already possess the HMAC key, so a second symmetric verification in `authorize` would not create an independent trust boundary and would reintroduce GitHub secret-tainting of job outputs.
- Added an inline workflow comment recording that invariant.
- Reduced retention from 14 days to 1.
- Added contract assertions for exact `contents: read` producer permissions, one-day retention, and absence of the HMAC key from every non-producer job.
- The existing `release-bus-build-profile.test.ts` proves the serialized profile is opaque (`JSON.stringify(profile)` contains none of the protected values) and that every protected value changes the HMAC digest.

Focused validation: 2 suites / 15 tests pass; Prettier passes. The earlier full frontend typecheck also passes.

PR #3415, opened 2026-07-22
Invoke frontend preflight shell gate directly

GelatoGenesis:
## Issue
- Frontend Release Bus preflight completed every gate and the immutable build, but aggregate invoked the shell gate through Node, causing a syntax failure before evidence was emitted.

## Fix
- Execute the staged shell gate directly for promotion-eligible fingerprint generation.

## Changes
- Remove the incorrect `node` prefix from the aggregate workflow step.
- Add a workflow contract assertion that requires direct execution and rejects the invalid invocation.

## Validation
- `./bin/6529 run test --testPathPatterns=__tests__/scripts/release-bus-frontend-gate.test.ts --maxWorkers=2` (13/13)
- `git diff --check`

## Risk
- Level: Low
- Why: one workflow command invocation and its regression contract; application code is unchanged.
- Rollback: revert this commit to restore the prior workflow invocation.

## Review Notes
- Focus on the executable boundary in the aggregate job; no frontend application deployment is required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release preflight evidence fingerprinting by invoking the release gate tool through its supported command interface.
  * Added safeguards to prevent incorrect tool invocation during promotion checks.

* **Tests**
  * Expanded preflight validation to verify the correct fingerprinting command and reject unsupported invocation patterns.

### DIFF
--- a/.github/workflows/release-bus-preflight.yml
+++ b/.github/workflows/release-bus-preflight.yml
@@ -665,7 +665,7 @@ jobs:
           contract="$RUNNER_TEMP/release-bus-preflight-contract.json"
           build_result="${{ needs.build.result }}"
           if [ "$PROMOTION_ELIGIBLE" = true ]; then
-            node "$RELEASE_BUS_GATE_TOOL" fingerprint \
+            "$RELEASE_BUS_GATE_TOOL" fingerprint \
               --repo-root "$GITHUB_WORKSPACE" \
               --base-sha "$SOURCE_SHA" \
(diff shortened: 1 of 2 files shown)


GelatoGenesis:
Addressed in ade1ff22c. The contract now proves the shell gate has a Bash shebang, is chmodded executable by `stage-tooling`, and is fail-closed by the downstream `Return aggregate result` step. The other two invocations intentionally retain `node` because they are `.cjs` tools. Focused workflow contract remains 13/13 green.

PR #3448, opened 2026-07-24
Fix staging immutable release cache exhaustion

GelatoGenesis:
## Issue
- Release Bus v2 staging train 29a4ee8c-aeb5-4975-827b-cb56135bf1d4 exhausted all three frontend deploy attempts because the managed instance root volume reached 100% usage.
- The immutable release directory retained every historical unpacked bundle, so future frontend trains would remain vulnerable even after one-off cleanup.

## Fix
- Bound the instance-side immutable release cache before downloading a new artifact.
- Preserve the exact healthy active rollback, the incoming release, and one newest additional rollback while pruning only validated one-level 40-hex release directories.
- Remove the transient package zip after its digest and extracted server have been verified.

## Changes
- Add fail-closed path and SHA validation around cache pruning in the Release Bus staging deploy workflow.
- Add workflow contract coverage for retention ordering, active/incoming preservation, path validation, and post-extraction zip cleanup.

## Validation
- `./bin/6529 run test __tests__/scripts/deployment-bus.test.ts --runInBand` — 60/60 passed.
- `./bin/6529 run check:changed` — changed-file typecheck and quality passed.
- `git diff --check` — clean.

## Risk
- Level: Medium
- Why: This changes staging instance cache lifecycle in a deployment workflow.
- Rollback: Revert this commit; active and incoming releases are never selected for pruning, and one extra rollback remains retained.

## Review Notes
- Focus on the exact-path safety assertions and pruning order after healthy pre-mutation verification but before artifact download.
- This does not enable Release Bus v2; both repository mode variables remain OFF and staging/production controls remain paused.

### DIFF
--- a/.github/workflows/release-bus-deploy-staging.yml
+++ b/.github/workflows/release-bus-deploy-staging.yml
@@ -261,6 +261,39 @@ jobs:
             }
           fi
 
+          prune_release_cache() {
+            current_release=''
+            if [[ "$previous_target" =~ ^${release_root}/releases/[a-f0-9]{40}/app$ ]]; then
+              current_release="${previous_target%/app}"
+            fi
+            retained_rollback=0
+            while IFS= read -r cached_release; do
+              [ -n "$cached_release" ] || continue
+              if [ "$cached_release" = "$release_dir" ] || \
+                [ "$cached_release" = "$current_release" ]; then
+                continue
+              fi
+              if [ "$retained_rollback" -eq 0 ]; then
+                retained_rollback=1
+                continue
+              fi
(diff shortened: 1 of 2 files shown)


GelatoGenesis:
Addressed the valid cache-entry findings in signed head 419d476f62b616f8d71a42110fc8cb56b3196f16: unrecognized directories are preserved and skipped instead of aborting all deploys; managed mode now fails closed unless the exact active release path is identified; the retention comment now distinguishes an already-cached incoming release. Contract coverage asserts the skip-before-delete behavior and managed-current guard. Focused workflow suite is 60/60. Please review the latest head.

PR #3458, opened 2026-07-24
fix: pin guarded outbound fetches

punk6529:
## Issue

The Open Graph preview path could validate a URL through the public-address guard and then perform the request through a separate caller-provided transport, bypassing the guard's DNS-pinned dispatcher.

## Fix

Keep transport ownership inside `fetchPublicUrl`. Callers may customize standard request options for each redirect hop, but cannot replace the validated dispatcher or network operation.

## Changes

- Replace custom fetch injection with a per-hop `RequestInit` builder.
- Preserve Open Graph host-specific headers and image-proxy HTTPS restrictions.
- Add regression coverage for initial and redirected DNS pinning, unsafe address families, legacy injection, dispatcher replacement, and concurrent batch fetches.

## Validation

- 61 focused Jest tests passed.
- Focused ESLint passed with zero warnings.
- Full strict TypeScript check passed.
- Full production build passed.
- Changed-file whitespace validation passed.

## Risk

Low-to-moderate. The transport is intentionally narrowed while request headers, redirect handling, timeouts, hostname/SNI behavior, and proxy protocol restrictions remain unchanged. The main compatibility risk is a caller that still attempts to supply the removed custom transport; repository-wide call-site review and regression tests cover that case.

## Review Notes

Please focus on redirect-hop request customization, DNS pinning ownership, and preservation of Facebook-specific headers and proxy HTTPS-only behavior.

## Summary by CodeRabbit

* **Security Improvements**
  * Strengthened protection against unsafe DNS resolutions, including private, link-local, loopback, and IPv6 addresses.
  * Improved security across redirect hops to prevent bypassing validated network destinations.
  * Ensured outbound requests retain validated connection settings and HTTPS restrictions.

* **Reliability**
  * Improved OpenGraph fetching behavior across HTML, file previews, metadata variants, and POST requests.
  * Added stronger concurrency handling for batches of OpenGraph requests.

### DIFF
--- a/__tests__/app/api/open-graph.route.test.ts
+++ b/__tests__/app/api/open-graph.route.test.ts
@@ -2048,9 +2048,9 @@ describe("open-graph API route", () => {
     );
     let activeFetches = 0;
     let maxActiveFetches = 0;
-    let releaseFetches = () => undefined;
+    let releaseFetches: () => void = () => undefined;
     const allFetchesStarted = new Promise<void>((resolve) => {
-      releaseFetches = resolve;
+      releaseFetches = () => resolve();
     });
 
     mockFetchPublicUrl.mockImplementation(
--- a/__tests__/app/api/open-graph.route.test.ts
+++ b/__tests__/app/api/open-graph.route.test.ts
@@ -131,6 +131,7 @@ const DEFAULT_USER_AGENT =
 
 const originalFetch = global.fetch;
 const mockFetch = jest.fn();
+const mockPreparedRequest = jest.fn();
 
 /**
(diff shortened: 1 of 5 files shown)


punk6529:
Addressed the GLM advisory concerns on the latest head. The implementation already reapplies redirect: manual and the internal timeout signal after the builder result, and toUndiciRequestInit allowlists fields while installing the pinned dispatcher itself. Commit 35227655d strengthens the regression test to prove caller- and builder-supplied dispatcher, agent, redirect, and signal values cannot replace those guard-controlled settings. The 61 focused tests, focused ESLint, source typecheck, and Jest test typecheck delta all pass.

PR #3469, opened 2026-07-26
FIX: Isolate public review artifacts by environment

punk6529:
## Issue
The Stream review UI and help corpus are intentionally local/staging-only. The dual-profile PR artifact currently builds its production profile with the job-level localhost endpoint, while staging rebuilds with `base-build` without regenerating environment-filtered public agent files. Release Bus v2 preflight likewise uses `base-build` directly.

## Fix
- Pin the PR artifact's production build to `https://6529.io`.
- Regenerate `help-index.json`, `glossary.json`, and `llms.txt` before the staging profile build.
- Regenerate the same environment-scoped artifacts for each Release Bus v2 build matrix profile.

## Validation
- `seize run test:no-coverage __tests__/scripts/deployment-bus.test.ts __tests__/scripts/release-bus-install-dependencies.test.ts`
- Production/staging help-corpus generation was exercised locally in the shell slice.

## Risk
Low and isolated to build-profile preparation. Production remains fail-closed; staging gains its intended review corpus. No deployment is triggered by this PR.

## Review notes
Please verify the production and staging artifact steps independently. This PR intentionally contains workflow-only changes so the GitHub App can write them with the required permission.


punk6529:
Resolved at afcc4922. The production step uses the package build script, whose prebuild hook already runs help-index:sync and agent-files:sync; the workflow now states that contract explicitly and runs the production, staging, and matrix profile blocks with fail-fast shell settings.

PR #3471, opened 2026-07-26
fix(public-review): normalize source commit timestamp

punk6529:
## Expected trust-root gate
The base-owned snapshot trust workflow fails intentionally because this PR changes the protected generator trust root. Its log requires an explicit maintainer bypass. The snapshot itself is not part of this PR; it will be regenerated and independently replayed after this fix merges.


punk6529:
Cross-platform replay evidence for exact head \21da165f39572dfaafed3256819ae8c802eb16eb\: the exported canonicalizer returns \2026-07-26T16:04:55Z\ for Stream commit seconds \1785081895\ under both the Windows host and \
ode:22.17.1-bookworm-slim\ on Linux. The protected snapshot-trust check is expected to fail on this generator trust-root change and its log correctly requires explicit maintainer bypass; the regenerated snapshot will receive the full base-owned trust replay after merge.

PR #3473, opened 2026-07-26
Validate Stream record-family review evidence

punk6529:
## Issue

- Stream now publishes typed record-family authorization evidence, but the deterministic public-review generator did not validate that evidence against the release manifest and pinned Solidity blobs.
- A selected semantic artifact could therefore be loaded without proving that its declared source bindings matched the exact review source.

## Fix

- Bind an optional record-family source catalog through its nested release-manifest record.
- Validate its source/candidate status, source blob checksums, authorization classes, family groups, host bindings, tests, and remaining blockers.
- Emit the validated catalog and a compact deterministic validation summary as auditor evidence.
- Keep older review configurations compatible when they predate this artifact.

## Notable Changes

- Add fail-closed record-family source-binding validation to the base-owned generator library.
- Add unit coverage for valid and drifted source hashes plus nested manifest binding.
- No review snapshot, route, runtime UI, or deployment behavior is changed in this PR.

## Validation

- Focused generator unit suite: 18 tests passed.
- The updated generator was also exercised against the pinned Stream `018c8788750980e143c38ace0666684bf641ec4f` corpus: 206 files, 420 definitions, 5,612 declarations, and all eight catalog source bindings validated.
- Changed-file lint and typecheck passed in the consuming snapshot worktree.
- Whitespace check passed.

## Risk

- Level: Medium.
- This changes a protected trust-root parser and must be merged before the data-only snapshot PR is accepted.
- Older configurations remain supported because the additional validation activates only when the artifact is selected.

## Review Notes

- Review the nested manifest lookup, exact checksum comparison, closed evidence-shape checks, and compatibility guard.
- The protected snapshot workflow is expected to require maintainer handling because this PR intentionally changes the base-owned generator; all other CI and bot feedback must be green first.

### DIFF
--- a/__tests__/scripts/public-review-solidity-reference.test.ts
+++ b/__tests__/scripts/public-review-solidity-reference.test.ts
@@ -924,6 +924,40 @@ describe("Solidity public-review reference generator", () => {
     expect(() =>
       validateRecordFamilyAuthorizationSourceCatalog(sources, artifacts)
     ).toThrow("source binding checksum drifted");
+
+    sources.set(sourcePath, { sha256: `sha256:${sourceSha256}` });
+    expect(() =>
+      validateRecordFamilyAuthorizationSourceCatalog(sources, {
+        "release-artifacts/record-family-authorization-source-catalog.json": {
+          ...catalogArtifact,
+          json: {
+            ...catalogArtifact.json,
+            family_groups: [
+              {
+                name: "ARTIST",
+                id: `0x${"c".repeat(64)}`,
+                allowed_authorization_class_ids: [2],
+              },
+            ],
+          },
+        },
+      })
+    ).toThrow("family evidence is malformed");
+
+    expect(() =>
+      validateRecordFamilyAuthorizationSourceCatalog(sources, {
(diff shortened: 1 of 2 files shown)


punk6529:
Addressed the actionable review feedback on current head `9be6ff2df`:

- family authorization-class IDs are now cross-referenced against the validated class set;
- malformed class records are rejected safely;
- negative fixtures now cover dangling class IDs, duplicate source bindings, duplicate class IDs, duplicate family IDs, and malformed family IDs.

I am retaining the existing two-phase manifest/catalog validation deliberately. `validateReleaseManifest` binds the manifest record to the exact artifact-map entry by path, SHA-256, byte size, and schema version. `buildBundle` then passes that same bound artifact entry to `validateRecordFamilyAuthorizationSourceCatalog`, which validates its semantics and hashes every source binding against the pinned Solidity source set. There is no second catalog or independently supplied digest that can diverge between those phases; merging semantic validation into the release-manifest structural validator would duplicate phase responsibilities and weaken backward compatibility for older manifests where this optional artifact is absent.

Focused Jest: 18/18 passing. Changed-file lint and whitespace checks pass locally.

PR #3484, opened 2026-07-27
Compose forward-only staging release commits

GelatoGenesis:
## Issue

- The frontend composition workflow could build the exact cumulative candidate tree, but it could not bind that tree to the current managed `1a-staging` head for a safe forward-only CAS.
- Forward-only rollback branches were also rejected by the workflow's immutable branch validator.

## Fix

- Accept an optional exact `release_parent_sha` and create a release commit whose first parent is that staging SHA and whose tree is the dependency-closed composition.
- Validate an existing immutable release branch against the exact first parent on retry.
- Accept immutable rollback branches so recovery can move shared staging forward to a restore commit instead of rewinding it.

## Changes

- Extend `release-bus-v2-compose.yml` with the release-parent input, first-parent validation, ancestor skipping, and deterministic `git commit-tree` release binding.
- Add an executable workflow regression proving the resulting commit contains cumulative A+B while first-parenting exact staging.
- Document CAS-before-deploy parity, durable crash recovery, and forward-only rollback in the durable operations guide.

## Validation

- `./bin/6529 run typecheck:jest`
- `./bin/6529 run test -- --runInBand __tests__/scripts/release-bus-v2-compose-workflow.test.ts` — 2 tests passed
- Prettier check for all changed files
- Composition workflow YAML parse
- `git diff --check`

## Risk

- Level: Medium
- Why: the application bundle is unchanged, but this expands an operational workflow contract used by the Release Bus controller.
- Rollback: pause `STAGING`, keep the backend controller undeployed or restore its prior version, and revert this workflow change. Existing callers that omit `release_parent_sha` retain the prior composition behavior.

## Review Notes

- Backend companion: 6529-Collections/6529seize-backend#1848. Merge this frontend workflow contract first.
- No frontend application deployment is required after merge.
- Focus on first-parent identity, immutable retry validation, DCO provenance, and rollback branch validation.
- This does not add a Reset Staging control or alter production qualification policy.

### DIFF
--- a/.github/workflows/release-bus-v2-compose.yml
+++ b/.github/workflows/release-bus-v2-compose.yml
@@ -79,8 +79,10 @@ jobs:
             git merge-base --is-ancestor "$BASE_SHA" "$existing"
             if [ -n "${RELEASE_PARENT_SHA:-}" ]; then
               git fetch --no-tags origin "$RELEASE_PARENT_SHA"
-              if [ "$existing" != "$RELEASE_PARENT_SHA" ]; then
-                test "$(git rev-parse "$existing^1")" = "$RELEASE_PARENT_SHA"
+              existing_parent="$(git rev-parse --verify "$existing^1" 2>/dev/null || true)"
+              if [ "$existing_parent" != "$RELEASE_PARENT_SHA" ]; then
+                echo "Existing release branch does not have the recorded staging ref as its first parent" >&2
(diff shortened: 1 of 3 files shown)


GelatoGenesis:
Superseded by #3485 on a fresh branch so the missing DCO trailer could be corrected without force-pushing shared history.

PR #3492, opened 2026-07-28
quality(create-wave): zero the typecheck-test debt + tighten with coverage

brookr:
## What & why

Structured, **enforced** quality improvements scoped to the create-wave feature, with **no runtime-behavior changes**. The area was already clean on the debt ratchet (0 `any_casts`/`test_generic_any`) and lint, so the real enforceable lever was the **Jest typecheck-test ratchet**, plus locking behavior with tests before the one small source tidy.

## Changes

**1. Zero the create-wave typecheck-test baseline (enforced ratchet ↓)**
- Resolved the **54 baselined Jest type diagnostics across 19 test files** by fixing the *real* mismatches — no `any` and no suppressions added:
  - optional-chained nested optional `period` accesses in the helper tests
  - `[n]!` on `noUncheckedIndexedAccess` array/query results (existing codebase idiom)
  - dropped unused `React` imports / locals
  - corrected the `ApiIdentity` fixture (numeric `cic`/`rep`, real classification enum) + `as unknown as` for deliberately-partial context/identity mocks
  - typed the p-limit / axios mocks without `any`
  - replaced the lone `@ts-ignore` with a typed cast
- Lowered `scripts/typecheck-test-baseline.json` by exactly those 19 entries (ratchet only permits decreases; committed in-PR).

**2. Narrow position-range parsing, drop non-null assertions (source)**
- `CreateWaveOutcomesManual.parseRange`: replaced 4 `!` with an explicit `undefined` guard + `Number.isNaN`. Behavior-identical.
- Added characterization tests first (reversed range `"3-1"`, start-below-one `"0-2"`) that pass against the pre-change source — locking the branches this touches.

**3. Characterization tests for the biggest untested branches**
- `Decisions`: disabling recurring cycles recomputes the end date from the last decision (toggle-off; toggle-on was already covered).
- `CreateWaveDraftsSection` (previously no spec): count badge, expand, load-and-collapse on selection, delete-by-id.

## Verification (local)
- `typecheck:jest` ratchet: **2152 → 2098** diagnostics; create-wave entries **19 → 0**.
- Debt ratchet: passes (no new `any`/TODO).
- `lint:changed` (tight): clean on all 24 changed files.
- Main app typecheck: zero create-wave errors.
- Jest: **97 suites / 599 tests pass** (TZ=UTC); new tests add zero typecheck debt.

No behavior changes — test-only except the behavior-preserving `parseRange` narrowing, which is guarded by the new characterization tests.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved winning-position validation with maximum limits and safer handling of invalid or reversed ranges.
  - Rolling waves now pin the end date to the last decision when rolling mode is disabled.

- **Improvements**
  - Added clearer validation feedback and accessible error announcements for manual outcomes.

- **Tests**
  - Expanded coverage for wave creation, drafts, dates, groups, outcomes, uploads, media handling, request formatting, modal behavior, and keyboard navigation.


brookr:
Thanks @coderabbitai — addressed both findings:

**🩺 Oversized rank positions → `RangeError` (fixed as a real bug).** Good catch. A winning position feeds `new Array(maxPosition)` at submit (and a wide range feeds `Array.from({ length })` while parsing), so a 10-digit rank crashed the form. Added a `MAX_WINNING_POSITION` (10,000) cap in `parseRange` that rejects oversized single values and ranges as invalid input before any allocation, plus regression tests for both (each throws pre-fix). — `3c62a97`

**📐 `as unknown as` suppressions.** Removed where a structurally valid form is practical:
- `CreateWaveNameInput`: ResizeObserver mock now has a real constructor + method signatures → valid `typeof ResizeObserver`, no cast.
- `CreateWaveOutcomesRowManual`: valid `CreateWaveOutcomeConfig` fixtures + typed mock-child props; assertions now check the real `type` field instead of a phantom `id` (which isn't on the type — the reason the old fixtures needed a double cast).
- `CreateWave`: ApiIdentity downgraded to a single `as ApiIdentity` (valid via structural overlap, matching sibling specs).

Kept deliberately, with reasons: the **p-limit/axios** doubles wrap third-party types (`LimitFunction`, and `AxiosResponse`'s opaque header/`InternalAxiosRequestConfig` shapes) that need a boundary cast even in a "complete" mock; and the **two context doubles** avoid 20+ field mocks the component never reads. Both are test-only boundaries, not production suppressions.

The `CreateWaveGroup`/helper-period cases are also test-only and behavior-preserving. Re-review welcome.

brookr:
Thanks @coderabbitai — all eight nits applied in `f7ae19cb5`:

**Maintainability**
- **Backdrop coupled to a Tailwind class** — added `data-testid="create-wave-modal-backdrop"` to the overlay in `CreateWaveModal.tsx` and the test now selects it by intent, so reshuffling utility classes can't silently break the dismissal test.
- **Calendar stub relied on `previousElementSibling`** — the stub's input now carries an id and the click handler looks it up directly, so the stub is order-independent.
- **Hardcoded winning-positions label** — switched all four new assertions to the suite's existing `getByLabelText(/Winning Positions/i)`.
- **`react-use` module replaced wholesale** — spread `jest.requireActual("react-use")` so only `useDebounce` is stubbed; any other hook added to this subtree keeps working instead of failing as `undefined`.
- **Raw DOM `.click()`** in `CreateWaveStepContent` → `userEvent.click`.
- **`URL.createObjectURL` reassignment** → `jest.spyOn(...).mockRestore()`.

**Correctness**
- **Test name promised more than it asserted** — the rejected-retry test now registers an `unhandledRejection` listener, awaits the settle, and asserts both that nothing went unhandled and that the error UI is still mounted. Renamed to match.
- **Mixed valid/invalid segments** — good catch, and you read the code right: `parsePositions` maps then `.filter(range => range !== null)`, so `"1,0-2"` and `"1,1-5000000000"` drop the rejected segment and submit only position 1 with **no** error. Pinned both as explicit characterization tests (`mixed valid and invalid segments (current behavior)`) rather than changing it — this PR is deliberately behavior-preserving, and tightening it to reject the whole input is a UX call for a follow-up. The tests make that trade-off visible so any future change to it is deliberate.

Suite is 805 tests / 103 suites green, Jest typecheck ratchet holds at 2096, debt ratchet unchanged, tight diff lint clean.

brookr:
Closing this — the work needs redoing rather than reviving, but **the underlying need stands** and the branch `quality/create-wave-ratchets` is being kept for reference.

## Why it can't just be refreshed

- **1069 commits behind `main`**, and a trial merge produces 7 conflicts, including an `add/add` on a test file someone else has since written their own version of.
- @ragnep landed *"Localize Create Wave outcome actions"* and *"Polish Create Wave responsive UI"* through the same two components, and @GelatoGenesis has since reworked `useWaveConfig`, group containment validation and step navigation. Resolving against all of that is close to redoing the work.
- Parts are already obsolete: @ragnep added aria wiring equivalent to this PR's WCAG fix, so that piece is handled upstream. The winning-positions fix has been extracted to **#3782** (merged-ready, verified on staging) and its predecessor #3541 is closed.
- `main` has re-accumulated **18 create-wave entries** in `scripts/typecheck-test-baseline.json`, so the debt this PR zeroed is back regardless.

## What was in here, worth redoing

- Zeroing create-wave's Jest typecheck debt (was 2152 → 2098 at the time; the ratchet is the only enforced lever for this area)
- ~140 new tests lifting create-wave coverage from 88.9%/78.3% to 93.9%/85.6% statements/branches, including first-ever specs for `CreateWaveInlineGroupXtdhGrant`, `CreateWaveInlineGroupRuleEditor`, `CreateWaveStepContent`, `CreateWaveModal` and `useMemeCardCount`
- Typed mocks replacing `(props: any)` throughout the specs it touched

## Known gap it surfaced and never fixed

`CreateWaveModal` has no `role="dialog"`/`aria-modal`, no focus trap and no Escape handling. Documented at the time, still true.

A fresh pass should be scoped against whatever `main` looks like then, and ideally coordinated so it does not run in parallel with active create-wave feature work.

PR #3505, opened 2026-07-28
test(public-review): decouple snapshot trust fixtures

punk6529:
## Issue
The snapshot trust test fixtures hard-coded `2026-07-28.1` as the next review version. Once that exact version is source-active in a candidate snapshot, the fixtures create duplicate retained/publication versions and fail before exercising the verifier behavior they are meant to test.

## Fix
- derive trusted history from the checked-out reference config
- append a fixed synthetic future review version that cannot collide with current retained history
- keep the existing immutable-history, publication-tamper, exact-base, cleanup, and regeneration assertions unchanged
- change tests only; production verifier and snapshot policy are untouched

## Validation
- snapshot trust suite: 41/41 passing
- changed-file lint: passing
- changed-file typecheck: passing
- Prettier: passing
- whitespace check: passing

## Risk
Low. One test file changes. The purpose is to keep the trust suite valid before and after real snapshot preloads without weakening any production trust invariant.

## Review notes
Please verify the synthetic future version is used only by fixtures and that trusted retained history still comes from the checked-out repository state.

## Summary by CodeRabbit

* **Tests**
  * Improved test consistency by using a shared synthetic future version across snapshot trust and publication validation scenarios.
  * Updated test setup to derive retained version history from configured values, improving resilience to configuration changes.
  * Preserved existing test behaviors and assertions while reducing hardcoded version data.

### DIFF
--- a/__tests__/scripts/public-review-snapshot-trust.test.ts
+++ b/__tests__/scripts/public-review-snapshot-trust.test.ts
@@ -165,6 +165,7 @@ const {
 const shaA = "a".repeat(40);
 const shaB = "b".repeat(40);
 const shaC = "c".repeat(40);
+const SYNTHETIC_FUTURE_VERSION = "2099-12-31.1";
 
 function loadConfig(): Config {
   return JSON.parse(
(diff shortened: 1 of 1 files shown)


punk6529:
/6529bot review general security

PR #3541, opened 2026-07-31
fix(create-wave): reject a winning-positions list with any invalid segment

brookr:
Follow-up to #3492, which pinned this behavior with characterization tests rather than changing it. Stacked on `quality/create-wave-ratchets` because it edits the same file; GitHub will retarget it to `main` when #3492 merges.

## The problem

`parsePositions` validated in two stages that disagreed with each other:

1. **Shape** — a regex over the whole string. Malformed input (`1--3`, trailing comma) rejected the **entire** input.
2. **Value** — each comma-separated segment went through `parseRange`, which returned `null` for a segment that was shape-valid but out of range (below 1, reversed, or above the 10,000 cap). Those `null`s were then **filtered out**, not treated as failures.

So a list mixing a valid position with an invalid one saved the valid part silently:

| Typed | Before | After |
|---|---|---|
| `1-3,5` | positions 1,2,3,5 | unchanged |
| `1--3` | "Invalid position format" | unchanged |
| `3-1` | "Invalid position format" | "Range 3-1 is backwards — put the lower position first" |
| `1,0-2` | **saves position 1, no error** | "Positions start at 1", nothing saved |
| `1,1-5000000000` | **saves position 1, no error** | "Positions can't go above 10,000", nothing saved |

A wave's winner structure is fixed at creation, so someone who fat-fingered one segment would only find out at announcement time that their outcome covered a fraction of what they intended.

Reviewers on #3492 converged on this independently — CodeRabbit flagged the filter, and 6529bot called it out as *"a genuine product trade-off… worth a follow-up product decision on whether a mixed list should reject rather than silently drop."*

## The change

- **Any invalid segment rejects the whole list**, matching how malformed input has always behaved.
- **Messages name the actual problem** instead of the generic "Invalid position format" — see the table above. "Invalid position format" is retained for genuine shape errors.
- Parsing moves to a module-level `parseWinningPositions` returning a discriminated result (`{ ok: true, positions } | { ok: false, error }`), which also drops the `try/catch` that existed only to swallow the oversized-allocation `RangeError` — segments are bounded before any array is built.
- The combined `(\d+(-\d+)?)(,\d+(-\d+)?)*` pattern is replaced by two flat per-segment patterns. The combined form nests quantifiers inside a repeated group and trips `security/detect-unsafe-regex`; per-segment validation is equivalent (empty segments, leading/trailing commas and `1-2-3` all still fail) and keeps the quantifiers as siblings.

## Tests

The two characterization tests from #3492 are flipped to assert rejection, plus new cases for a backwards segment later in the list, first-offender message ordering when several segments are bad, both sides of the 10,000 boundary, and that an all-valid list still dedupes and sorts (`3,1-2,2` → 1,2,3). 30 tests in this spec, 810 across create-wave.

## Verification

- create-wave suite: 103 suites / 810 tests green
- Jest typecheck ratchet: 2096 in 855 files (no new debt)
- Debt ratchet: unchanged
- Tight diff lint: clean, no suppressions
- React Doctor: 100/100

## Noted, not addressed

This component hardcodes all of its copy ("Manual action", "Save", "Please enter positions"), so the new strings follow suit rather than half-migrating the file to `i18n`. An i18n pass over the whole component is separate work.

Brook also wants a prod usage probe before we invest further here — the suspicion is that nobody configures anything beyond top 1 / top 3 / top 5, in which case this field is far more configurable than it needs to be. That's queued separately, not part of this PR.


brookr:
Closing in favour of **#3782**, which carries this exact fix rebuilt on current `main`.

Why replaced rather than refreshed:

- This PR was stacked on #3492, which has since fallen **1069 commits** behind `main` and now conflicts with @ragnep's Create Wave localization and responsive work in the same two components.
- `main` has moved underneath it: the component is now localized (`useBrowserLocale`/`t()`), and @ragnep added aria wiring equivalent to part of what #3492 contributed — so that piece is already handled upstream and correctly dropped.
- CI has moved too. `App check / Quality and contracts` with a Knip dead-code step arrived with #3593 and did not exist when this branch was written; the code as it stands here would fail it.

#3782 is the validation fix alone, applied on top of `main` as it stands today: 25 tests in that file (9 of which fail against the current parser), full gates green, and verified on staging at `d451b26f71`.

No functionality is lost by closing this.

PR #3543, opened 2026-07-31
Make NFT media E2E assertions type-agnostic

prxt6529:
## Issue

- The read-only `/the-memes/mint` staging E2E assumed the latest mint rendered an image or iframe, so it failed when The Memes #529 used video media.

## Fix

- Add one stable `data-nft-media-renderer` marker to the actual rendered NFT media element for image, video, HTML, and GLB media.
- Make the marker value identify the renderer type while allowing E2E coverage to use one type-agnostic locator.

## Changes

- Mark the real `img`, `video`, `iframe`, and `model-viewer` elements.
- Update the mint-page E2E to assert a visible marked renderer without assuming live media type.
- Extend focused renderer tests to enforce the shared marker contract for all four supported media types.
- Leave renderer-specific behavior and selectors intact; no other equivalent incomplete E2E assertion was found.

## Validation

- Focused NFT image, video, HTML, and GLB renderer tests.
- Changed-file lint.
- Production TypeScript typecheck.
- React Doctor diff review.
- Frontend debt ratchet.
- Playwright typecheck was not completed locally because the installed dependency set lacks `@axe-core/playwright`; CI was not awaited per request.
- Local Playwright runtime E2E was not run because it starts a development server, which was outside this task.

## Risk

- Level: Low
- Why: Adds inert data attributes to existing media elements and narrows the E2E assertion to a shared renderer contract.
- Rollback: Revert this PR to restore the prior selectors and markup.

## Review Notes

- Confirm the marker remains on each actual media element rather than an outer layout container.

## Summary by CodeRabbit

* **Enhancements**
  * Added consistent media-renderer markers for NFT images, videos, HTML content, and 3D models.
  * Improved media identification across mint-page displays.

* **Tests**
  * Added coverage confirming renderer markers are present and correctly assigned.
  * Updated mint-page media checks to use the shared renderer marker.


prxt6529:
Follow-up triage:

- Fixed the task-owned Knip failure in `471dce2df` by keeping the helper's attribute constant and media-type union private.
- The GLM advisory that video lacks the marker is stale: `NFTVideoRenderer` already passes the marker through the real `SeizeVideoPlayer`, and the focused video test exercises the resulting `<video>`.
- I did not increase the E2E timeout because this task explicitly requires a renderer-contract fix rather than a timeout workaround.
- The HTML marker intentionally exists only when an actual iframe source renders; the GLB assertion targets the actual `model-viewer`; and the durable PR report is required by the repository workflow.
- SonarCloud currently reports zero new-code issues and an OK quality gate, so its earlier failure comment is stale.

PR #3546, opened 2026-08-01
fix(composer): stop losing @mentions on draft restore and markdown import

brookr:
## The problem

**Tagging someone in a wave composer can silently fail to tag them.** The drop posts, the text looks fine while you are typing it, but the sent message shows `@[handle]` as plain unclickable text and the person is never notified.

It is easy to hit and hard to notice: the mention looks completely normal in the composer right up until you send, and the author gets no error. From the reader's side it just looks like a typo. Anyone relying on a tag to reach someone — reviews, approvals, hand-offs — can be silently dropped.

The stored drop ends up with the mention token still in `content` but no corresponding entry in `mentioned_users`, which is what the renderer keys links off, so it prints the raw token.

Two independent defects cause this.

## 1. Mentions are lost when a draft is restored

The wave composer autosaves a draft and restores it on reload. That draft persists `editorState.toJSON()` and nothing else (`useWaveDraftPersistence.ts:65`).

Mention nodes survive that round trip intact — `exportJSON`/`importJSON` both carry `mentionedProfileId`. What does **not** survive is `mentionedUsersRef`, the list of profiles picked from the autocomplete during this session, which is what submission actually reads. It is only ever repopulated by `restoreMentionedEntities` on a failed-send retry, never on draft restore.

So after a reload the editor shows correct, styled mentions while the registry behind them is empty, and the drop submits with none of them. A mention added *after* the restore registers normally — which is why a single message can contain a mix of working and broken tags.

**Repro:** open a wave composer, tag two or three people from the autocomplete, wait ~1s for the autosave, reload the page, send. Every restored mention posts as dead text. Confirmed manually.

**Fix:** `getMentionedUsersFromEditorState` reads mentions straight off the editor nodes — mirroring the existing `getMentionedGroupsFromEditorState` — and is merged over the session registry so alias expansion (which registers through `onMentionedUser`) keeps working. The editor is the source of truth because it is the thing that survives a restore.

## 2. Markdown import only converted a token at the very start of the text

`@lexical/markdown` matches text-match transformers and splits the text node at the match offset:

```js
const match = textNode.getTextContent().match(transformer.importRegExp);
const startIndex = match.index || 0;
```

With a `g` flag, `String.prototype.match` returns **every** match and `index` is `undefined` — so `startIndex` silently fell back to `0` and the split landed in the wrong place. Only a token at position 0 converted; everything else was left as literal markup.

| Input | Before | After |
|---|---|---|
| `hey @[bob] take a look` | nothing converts | converts |
| `@[alice] @[bob] @[carol] …` | only `@alice` | all three |
| `check #[Some Wave] later` | nothing converts | converts |
| `look at $[card] here` | nothing converts | converts |

This affected user mentions, wave mentions, NFT references and emoji alike, on every path that re-imports content — including opening a drop to edit it.

**Fix:** drop the `g` flag from the affected transformers. `EMOJI_MATCH_REGEX` keeps it because `EmojiPlugin` drives it with `matchAll`; the transformer gets a non-global twin.

## Tests

22 new tests, driving the **real** Lexical engine rather than the repo's `lexical` mock:

- `markdownImportTokenSplitting.test.ts` (12) — every token type, at the start of the text, mid-text, and several per line, plus flag assertions. **10 of the 12 fail against the previous regexes**, verified by stashing the fix and re-running.
- `userMentionDetection.test.ts` (10) — extraction, skipping nodes with no profile id, dedupe, merge precedence, and the `toJSON()` → `parseEditorState` draft round trip that reproduces the original bug end to end.

Why the existing suite missed all of this: `MentionTransformer.test.ts` stubs `$createMentionNode` as `(text) => ({type, text})`, discarding the profile id entirely, and every suite mocks `lexical` wholesale — so no test ever exercised a real markdown import.

## Verification

- Full suite: 2125 suites / 13,028 tests green
- Jest typecheck ratchet: 2140 in 872 files
- Debt ratchet: unchanged
- Tight diff lint: clean, no suppressions
- React Doctor: 99/100 — both warnings are pre-existing size and `useState` counts on `CreateDropContent`; this PR adds a `useMemo` and no new state

## Note on touched tests

Three existing suites needed a line each. Two feed a stub editor state and already mock the sibling group-mention reader, so they now mock the user-mention one the same way; the alias integration test passes the new prop explicitly. No assertion was weakened and no behaviour was changed to make tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed mention, wave mention, hashtag, emoji, and NFT reference parsing within surrounding text and repeated sequences.
  * Preserved accurate token positions during Markdown import.
  * Improved restoration of mentions, wave identities, and NFT references from saved draft content.
  * Ensured current editor metadata takes precedence while retaining registry-only entries.

* **Tests**
  * Added coverage for metadata persistence, deduplication, conflicts, draft submissions, and token parsing.


brookr:
Thanks @coderabbitai — all three applied in `6daed1bdd`, though one landed differently than suggested:

- **Share one regex constant per transformer** — done. `MENTION_MATCH_REGEX`, `WAVE_MENTION_MATCH_REGEX` and `NFT_REFERENCE_MATCH_REGEX` are each a single named constant used for both `regExp` and `importRegExp`, so the two literals can no longer drift apart. Safe to share the same object here precisely because they are non-global — a `lastIndex` would otherwise be shared state across the two call sites.

- **Cover the emoji tokenizer** — done, two cases plus a flag assertion. Emoji needed `$nodesOfType` rather than the text-node walk the other three use, since `EmojiNode` is a `DecoratorNode`.

- **Derive the non-global emoji regex instead of duplicating the pattern** — tried it, had to back it out. `new RegExp(EMOJI_MATCH_REGEX.source)` trips this repo's `security/detect-non-literal-regexp`, and suppressing a security rule for cosmetics isn't a trade I want to make. There's also a second-order problem: computing at module scope means a test that mocks `EmojiPlugin` crashes on `.source` of `undefined` at import time, which is exactly how I hit it the first time round.

  Your underlying concern is right, so it's addressed a different way — the literal is back with a comment explaining why, and a test now asserts `EMOJI_SINGLE_MATCH_REGEX.source === EMOJI_MATCH_REGEX.source` along with their flags. Drift becomes a failing test rather than a silent divergence, which is a stronger guarantee than deriving would have given.

Suite is 2127 suites / 13,038 tests green, typecheck ratchet 2140, debt ratchet unchanged, tight lint clean with no suppressions. Re-verified that the tokenizer tests still bite: 13 of the 15 fail against main's transformers.

PR #3558, opened 2026-08-03
Fix Poper Blocker AWS RUM fetch rejection filter

simo6529:
## Issue

- The client Sentry filter for Poper Blocker's orphan fetch rejection misses the current production shape because the known injected frame at line 1, column 4520 is unsymbolicated and has no function label.
- That causes an extension-induced AWS RUM transport rejection to remain in product error telemetry even though the exact injected path, coordinate pair, network TypeError, and unhandled-rejection mechanism are otherwise unchanged.

## Fix

- Allow only the proven first `app:///injectScriptAdjust.js` frame at `1:4520` to use either `window.fetch` or a null/absent function label.
- Continue requiring the exact `VihJ` frame at `1:3159`, exactly two matching injected frames, one network-failure `TypeError`, the browser `onunhandledrejection` mechanism, `handled: false`, and no application-owned source evidence.

## Changes

- Narrowly extended the existing Poper Blocker frame comparator.
- Added filter-level and `beforeSend` regressions for the current production shape.
- Added strict near-misses for an empty or incorrect function, an unsymbolicated second frame, a duplicate first signature without `VihJ`, a nearby script path, AbortError, HTTP failure, timeout, handled frameless network failure, and existing app-owned or mixed stacks.
- Kept the monitoring relationship represented only by the public `aws-rum-web` package frame; no endpoint, AppMonitor identifier, profile value, or raw event identifier is retained.

## Validation

- `./bin/6529 run test __tests__/utils/sentry-client-filters.test.ts __tests__/instrumentation-client.test.ts` — 2 suites and 423 tests passed.
- `./bin/6529 run lint:changed` — passed.
- `./bin/6529 run typecheck:ci` — passed.
- `./bin/6529 run check:changed` — changed-source typecheck and lint passed; the aggregate command then failed on the repository's existing unrelated Knip unused-file/export inventory.
- Scoped diff/privacy scan — no raw monitor IDs, UUIDs, wallet/profile data, private URLs, or local paths.
- Browser runtime was not run locally because this is a pure pre-send predicate with no UI, AWS/Sentry configuration, or application request behavior change. The PR CI-selected Playwright smoke pack passed on the initial code head and will rerun for the latest test-only head.

## Risk

- Level: Low
- Why: The accommodation applies to one coordinate-anchored frame only; every other existing ownership, mechanism, message, frame-pair, and fail-open guard remains intact.
- Rollback: Revert the signed commits to restore the previous exact function-label requirement.

## Review Notes

- Please focus on whether the first-frame null/absent allowance remains sufficiently narrow and whether the near-miss coverage proves genuine application and network failures stay reportable.
- This may overlap conceptually with the separate Sentry path-privacy task, but it does not depend on that work or change path sanitization.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error filtering for browser extension and AWS RUM fetch-related failures.
  - Reduced false-positive error reports from blocked or unsymbolicated Poper Blocker requests.
  - Preserved reporting for legitimate, handled network failures, including cases without stack frames.
  - Improved handling of incomplete, malformed, or partially matched stack-frame information to prevent unrelated errors from being filtered incorrectly.


simo6529:
@coderabbitai review

simo6529:
@coderabbitai review

PR #3590, opened 2026-08-04
Remove duplicate GDRC document header

ragnep:
## Issue

- The GDRC page rendered its title and version in the React header while the fetched charter HTML retained the same leading header in the DOM.

## Fix

- Remove only the exact leading fetched title/version block after the existing HTML sanitization step.
- Keep the React-rendered title/version as the single page header and preserve all charter wording.

## Changes

- Match the known source header structure and remove it only when both strings exactly match the localized page header.
- Make the Preamble the first charter section instead of hiding the first fetched block with CSS.
- Add regression coverage confirming there is one title, one version, and the charter content remains present.

## Validation

- `./bin/6529 run test --testPathPatterns=__tests__/components/about/AboutGDRC1.test.tsx --runInBand --cacheDirectory=/private/tmp/6529-gdrc-jest-cache`
- `./bin/6529 run react-doctor:diff` (100/100)
- Browser checked `/about/gdrc1` at desktop and 390 x 844: one title, one version, Preamble and Digital Rights preserved, no horizontal overflow, and no new console errors.

## Risk

- Level: Low
- Why: The change is isolated to the GDRC fetched-content sanitizer, its local CSS, and focused tests. Removal requires an exact structure and exact title/version text match.
- Rollback: Revert the single commit to restore the previous CSS-hidden source header behavior.

## Review Notes

- Please focus on the narrow duplicate-header predicate and preservation of the remaining sanitized charter HTML.
- No documentation or Help index update is needed because the route, terminology, controls, workflow, and charter wording are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicated charter headers appearing in GDRC content.
  * Ensured charter content, version information, and the Preamble heading remain visible.
  * Improved content updates when translated title or version text changes.

* **Tests**
  * Added coverage confirming duplicate headers are removed without hiding valid charter content.

### DIFF
--- a/__tests__/components/about/AboutGDRC1.test.tsx
+++ b/__tests__/components/about/AboutGDRC1.test.tsx
@@ -29,6 +29,25 @@ describe("AboutGDRC1", () => {
     expect(await screen.findByText("Charter Content")).toBeInTheDocument();
   });
 
+  it("removes the duplicated source header without removing charter content", async () => {
+    (fetchAboutSectionFile as jest.Mock).mockResolvedValue(
+      "<div><h2>The Global Digital Rights Charter</h2><div>Version 1.0 - March 2023</div></div><div><h3>Preamble</h3><p>Charter Content</p></div>"
+    );
+
+    render(<AboutGDRC1 />);
+
+    expect(await screen.findByText("Charter Content")).toBeInTheDocument();
+    expect(
+      screen.getAllByRole("heading", {
+        name: "The Global Digital Rights Charter",
+      })
+    ).toHaveLength(1);
+    expect(screen.getAllByText("Version 1.0 - March 2023")).toHaveLength(1);
+    expect(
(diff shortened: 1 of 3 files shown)


ragnep:
Addressed the valid i18n/robustness finding in `071274496`:

- duplicate detection now matches the fixed fetched document against the canonical source-locale title/version, independent of the active display locale;
- the effect is back to a single initial fetch;
- targeted Jest, changed-file lint/typecheck, and React Doctor all pass again.

The `textContent` nullability finding does not apply to these `Element` values in the repo's DOM typings: direct `.trim()` passes strict typecheck, while optional chaining is rejected by the configured lint rule as unnecessary. The direct-child check is intentional so a source block containing any additional text is never removed. The first-element requirement is also intentional and matches the verified public source structure; broadening it could remove content outside the known duplicate header.

PR #3597, opened 2026-08-04
Run staging Git sync as the checkout owner

punk6529:
## Outcome\n\nRepairs the first live accelerated staging promotion without weakening Git's ownership protection.\n\nThe exact staging artifact built successfully in 11m05s. SSM then ran repository synchronization as root against an ubuntu-owned checkout, so Git failed closed with dubious ownership before any activation. This patch:\n\n- runs every fetch/rev-parse/reset/clean/checkout/config operation as the configured checkout owner;\n- retains root only for the controlled artifact installation steps that require it;\n- explicitly rejects a global safe.directory bypass in the workflow contract;\n- records the exact production/staging benchmark and failure disposition.\n\n## Evidence\n\n- failed staging run: 30959063209\n- SSM command failed before activation; production was unaffected\n- workflow parses and owner-bound Git contract passes\n- changed-workflow security validation passes\n- portable diff check passes\n- hosted exact-head App CI is authoritative for the full Linux workflow contract

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -592,21 +592,29 @@ jobs:
             fi
           done
 
-          cd "$REPO_DIR"
-          git fetch --no-tags origin "$BRANCH"
-          remote_sha="$(git rev-parse FETCH_HEAD)"
+          # SSM runs this wrapper as root, while the checkout and its Git
+          # credentials belong to RUN_AS. Keep every repository operation
+          # under that identity; do not weaken Git's ownership protection with
+          # a root-level safe.directory exception.
+          run_git() {
+            sudo -H -u "$RUN_AS" git -C "$REPO_DIR" "$@"
+          }
+
+          run_git fetch --no-tags origin "$BRANCH"
+          remote_sha="$(run_git rev-parse FETCH_HEAD)"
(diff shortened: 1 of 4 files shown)


punk6529:
/6529bot review general security

punk6529:
Exact-head review disposition for c6954eb2c8d63272b59df215255f0cd39b6aaea:\n\n- 6529bot general: Good to merge.\n- 6529bot security: No findings.\n- The deploy script is intentionally entered as root after repository sync; its existing contract owns privilege drops for PM2 and checkout-user files while root owns controlled installation/rollback.\n- The suggested extra reset/clean string assertions are safe follow-up hardening, but the current contract already asserts the owner-bound 
un_git implementation, owner-bound fetch/checkout, no global ownership bypass, and the full exact-head Linux Release Bus contract. Restarting an 11-minute production build for duplicate string assertions would add no runtime confidence to this urgent failed-before-mutation repair.\n- All required/advisory checks are green; no review threads are open; base and head are exact.\n\nThe remaining BLOCKED state is the team-review/last-push rule. Under the repository owner's explicit AFK authorization, I am applying the maintainer/admin bypass to this exact reviewed hotfix so staging qualification can resume.

PR #3706, opened 2026-08-10
feat: auth-bridge page for community app authentication

RegularDad6529:
### What

Adds a new page at `/auth-bridge` that lets community-built applications authenticate 6529.io members with zero friction — no second wallet signature needed.

### How it works

1. Community app (e.g. AR app on Arweave) links to `6529.io/auth-bridge?r=https://arweave.net/XXXX`
2. Page reads the existing JWT from the user's cookie/localStorage via `getAuthJwt()`
3. Redirects back to the app with the token in the URL **fragment** (`#t=JWT`)
4. The receiving app reads the token from the fragment, stores it, and clears the fragment

If the user is already logged into 6529.io, the entire flow is an instant flash — no wallet popup, no extra clicks.

### Security

- **Allowlist-based redirect validation** — only origins in `AUTH_BRIDGE_ALLOWLIST` are accepted (no open redirect attacks)
- Currently allows `https://arweave.net` for community AR apps
- New apps can be added to the allowlist array
- Token in URL **fragment** (`#t=...`), not query string — never hits server logs, purely client-side
- If user not logged in, shows an error message and redirects to 6529.io home

### Why

Enables community projects like the 6529 AR platform (world-anchored AR layer gated by 6529.io identity) and other tools to piggyback on 6529.io's wallet sign-in without building separate auth infrastructure or asking users to sign again.

### Files changed

- `app/auth-bridge/page.tsx` — server component (metadata + Suspense wrapper)
- `app/auth-bridge/page.client.tsx` — client component (reads JWT, validates redirect, redirects with token)

### Usage example

```text
// Community app on Arweave:
// 1. User clicks: https://6529.io/auth-bridge?r=https://arweave.net/my-ar-app
// 2. Instant redirect back: https://arweave.net/my-ar-app#t=eyJhbGci...
// 3. App reads fragment, stores JWT in localStorage, clears fragment
```

## Summary by CodeRabbit

* **New Features**
  * Added an authentication bridge page for redirecting authenticated users to approved destinations.
  * Added strict validation for redirect targets to help prevent unauthorized redirects.
  * Added clear loading, redirecting, and error states during authentication.
  * Unauthenticated visitors are redirected home after a brief delay.
  * Added configurable approved callback addresses and page metadata for the authentication bridge.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -28,6 +28,10 @@ TEST_WORKERS=2 # Suggested: half the number of CPU cores
 # MOBILE APP SCHEME
 MOBILE_APP_SCHEME=
 
+# AUTH BRIDGE — comma-separated exact callback URIs for community app auth
+# Example: NEXT_PUBLIC_AUTH_BRIDGE_ALLOWLIST=https:...ck
+NEXT_PUBLIC_AUTH_BRIDGE_ALLOWLIST=
+
 # 6529 CORE SCHEME
 CORE_SCHEME=core6529
 
--- a/app/auth-bridge/page.client.tsx
+++ b/app/auth-bridge/page.client.tsx
@@ -1,0 +1,121 @@
+"use client";
+
+import { useEffect, useState } from "react";
+import { useSearchParams } from "next/navigation";
+import { getAuthJwt } from "@/services/auth/auth.utils";
+
+/**
+ * Auth Bridge Page
+ *
+ * Enables community-built applications (e.g. AR apps on Arweave) to
+ * authenticate 6529.io members with zero friction. If the user is already
+ * logged into 6529.io, this page reads the existing JWT and redirects back
+ * to the requesting app with the token in the URL fragment (#t=...).
+ *
+ * The token is placed in the fragment (not the query string) so it never
+ * hits server logs — it is purely client-side.
+ *
+ * Redirect targets are validated against an allowlist to prevent open
+ * redirect attacks. New community apps can be added to AUTH_BRIDGE_ALLOWLIST.
+ */
+
+// Allowlist of approved redirect targets.
+// To add a new community app, add its origin here.
+const AUTH_BRIDGE_ALLOWLIST: string[] = [
+  "https://arweave.net",
+  // Add more approved community app origins here
+];
+
+function isAllowedRedirect(target: string): boolean {
+  try {
+    const url = new URL(target);
+    return AUTH_BRIDGE_ALLOWLIST.some(
+      (allowed) => url.origin === allowed
+    );
+  } catch {
+    return false;
+  }
+}
+
+export default function AuthBridgePageClient() {
+  const searchParams = useSearchParams();
+  const [status, setStatus] = useState<"loading" | "error" | "done">(
+    "loading"
+  );
+  const [message, setMessage] = useState<string>("");
+
+  useEffect(() => {
+    const redirectTarget = searchParams.get("r");
+
+    if (!redirectTarget) {
+      setStatus("error");
+      setMessage("Missing redirect target.");
+      return;
+    }
+
+    if (!isAllowedRedirect(redirectTarget)) {
+      setStatus("error");
+      setMessage(
+        "This redirect target is not approved. Contact the 6529 team to add your app to the allowlist."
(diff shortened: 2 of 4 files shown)


GelatoGenesis:
**Blocking security concern: this forwards the user’s full 6529 bearer credential to another origin.**

I checked the current auth flow, and this would mechanically work. The durable `6529_session` cookie is HTTP-only, but the access JWT returned by `getAuthJwt()` is not: it is stored in the JavaScript-readable `wallet-auth` cookie and also retained in localStorage. Therefore `/auth-bridge` can read it and place it in the destination fragment.

The receiving application can then use that JWT from a non-6529 domain by sending `Authorization: Bearer <JWT>`. Our general API CORS policy allows arbitrary origins and the `Authorization` header, and access-token verification is not bound to the browser’s `Origin`. In other words, possession of this JWT is sufficient to use it.

That is the main problem: this is not a limited identity assertion. It hands the external application the user’s normal, unscoped 6529 access token, allowing it—or anyone who obtains the token—to perform any authenticated action available to that user until the token expires.

Using a URL fragment only prevents the token from being included in the initial HTTP request/server logs. It does not protect it from the destination application’s JavaScript, XSS, dependencies, analytics, browser extensions, or accidental persistence.

The Arweave case is especially concerning. If the callback is `https://arweave.net/<transaction>`, browser storage is scoped to the `https://arweave.net` origin, not to an individual transaction path. If the receiving application stores the JWT in localStorage as proposed, arbitrary content hosted at another `https://arweave.net/...` path is potentially in the same storage origin. An exact callback-URI allowlist does not provide storage isolation.

There are also functional gaps:

- The allowlist defaults to empty, and this PR does not appear to inject `NEXT_PUBLIC_AUTH_BRIDGE_ALLOWLIST` into the production build, so the submitted version would reject every callback unless deployment configuration is changed separately.
- `getAuthJwt()` can fall back to an expired JWT in localStorage. The bridge does not validate expiry or wait for the normal HTTP-only session refresh, so a legitimately logged-in user could be redirected with an unusable token.
- The external application cannot use the first-party HTTP-only session cookie to refresh the exported token, because session-management endpoints are origin restricted.

I do not think we should transfer the primary 6529 JWT this way. A safer design would use:

1. A dedicated, application-controlled origin rather than the shared `arweave.net` origin.
2. A short-lived, one-time authorization code tied to the exact callback, with `state` and PKCE.
3. A backend exchange that returns a narrowly scoped, audience-bound token containing only the permissions/identity information the community application needs.

That would provide the desired low-friction sign-in without giving a third-party application the user’s normal 6529 session authority.

RegularDad6529:
Thanks for the thorough review, @GelatoGenesis — you were right that forwarding the raw JWT to a third-party app is fundamentally insecure.

I've redesigned this as a **QR-based device authorization flow** (RFC 8628-inspired) that reuses 6529.io's existing connection-share infrastructure but changes what token gets minted at the end.

**The new flow:**
1. Community app generates PKCE pair, redirects to `6529.io/auth/device`
2. 6529.io shows QR code (existing QR generation)
3. User scans on phone → sees app name + requested scopes → taps Authorize
4. Community app exchanges device code + PKCE verifier for a **scoped, audience-bound** JWT

**How it addresses your concerns:**
- **No raw JWT exposure** — community apps never see the user's 6529 session token
- **Scoped tokens** — `type: 'community_app'`, `aud: <client_id>`, `scope: 'identity:read'` — cannot be used as normal 6529 session tokens
- **PKCE S256** — prevents token interception even if redirect URI is compromised
- **15-minute TTL, no refresh** — re-scan QR when expired
- **Explicit user consent** — user sees app name + scope and must tap Authorize

Closing this PR in favor of the new approach:
- **Frontend:** #3938
- **Backend:** 6529-Collections/6529seize-backend#1990

The backend PR adds the community app registry, device code tables, scoped JWT issuance, and three new API endpoints. The frontend PR adds the `/auth/device` QR page and `/auth/device/approve` mobile approval page.

RegularDad6529:
Closing in favor of QR-based device authorization flow. See #3938 (frontend) and 6529-Collections/6529seize-backend#1990 (backend).

PR #3761, opened 2026-08-18
Refine upcoming Meme transitions

prxt6529:
## What changed

- reveal the explicitly mapped canonical next Meme on its detail page without waiting for the current mint to complete
- keep the homepage on `Latest Drop` until ten minutes after the shared scheduled mint-end timestamp, then replace it with `Next Drop`
- expose the homepage grace period through `HOME_LATEST_DROP_GRACE_PERIOD_MINUTES`
- update focused coverage, product documentation, and the Help Bot index

## Why

The upcoming Meme detail page was incorrectly coupled to the homepage transition condition. The homepage also switched to the next winner immediately when the current mint completed, without a short closing window for the latest drop.

## Impact

Visitors can view and subscribe to the mapped upcoming Meme immediately from its detail route. The homepage retains its existing `Latest Drop` and `Next Drop` labels and switches between them at a consistent configured time.

## Validation

- focused Jest coverage for upcoming Meme reveal and homepage transition boundaries
- changed-file type-check, formatting, lint, and quality checks

## Summary by CodeRabbit

* **New Features**
  * Latest Drop now remains visible for a configurable grace period after minting ends before switching to Next Drop.
  * Upcoming Meme pages now show loading, reveal, and winner states more consistently.
  * Canonical upcoming Meme routes can display the winning artwork panel regardless of homepage mode.

* **Documentation**
  * Updated help content and feature documentation to explain grace-period timing and upcoming Meme artwork behavior.


prxt6529:
Follow-up addressed in `98f9ecfd6`: added regression coverage proving transition readiness is re-keyed when the mint number changes and remains false when the mint is unavailable. The existing `elapsedTransitionTime === transitionTime` guard already prevents stale readiness, so no state reset is needed. Updated the documentation to describe the scheduled mint-end deadline without promising clock-perfect client rendering. Confirmed `mintEndUtc` is a required `Date` returned by `getMintTimelineDetails`, so the suggested undefined crash path is not possible. The upcoming Meme detail page remaining independent of the active mint is intentional product behavior.

PR #3808, opened 2026-08-21
Fix NFT reference card hydration markup

prxt6529:
## Issue

- NFT reference cards are rendered inside Markdown paragraphs but contain div elements, producing invalid paragraph, anchor, and div nesting plus React hydration warnings.

## Fix

- Render the NFT card wrappers and details row with phrasing-content elements while preserving the existing flex layout and link behavior.

## Changes

- Replace the NFT card wrapper divs with spans.
- Update the card element ref type to match the rendered element.
- Add regression coverage that renders the complete NFT card inside a paragraph and asserts there are no nested div elements.

## Validation

- ./bin/6529 run test --testPathPatterns=__tests__/components/drops/view/item/content/nft-tag/DropListItemContentNft.test.tsx --runInBand
- ./bin/6529 run lint:changed
- ./bin/6529 run typecheck:changed
- ./bin/6529 run react-doctor:diff

## Risk

- Level: Low
- Why: The change only substitutes semantic wrapper elements while retaining the existing classes, links, images, and data flow.
- Rollback: Revert the single fix commit.

## Review Notes

- React Doctor scored the diff 99/100. Its two warnings concern pre-existing img elements and are unrelated to the hydration fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NFT content markup for better compatibility with inline text and phrasing content.
  * Updated decorative NFT images to use empty alternative text, preventing redundant screen reader announcements.
  * Preserved NFT image and token-name display while handling missing image data correctly.


prxt6529:
Addressed the i18n/accessibility finding in 790f1cfff. Both NFT images are decorative within a link whose visible token name already supplies the accessible name, so their alt attributes are now empty. Focused tests assert the decorative image semantics.

PR #3821, opened 2026-08-24
Make Broadcast mentions control @all and @contributors

prxt6529:
## Issue

- `@contributors` is a broad Wave broadcast but the composer offered it to ordinary chat participants.
- The notification setting was labeled ALL mentions even though the same preference should control both admin-only broadcasts.

## Fix

- Apply the existing creator/admin mention permission to both `@all` and `@contributors` while leaving escalation mentions available to chat participants.
- Rename the recipient setting to Broadcast mentions and define it as the preference for both `@all` and `@contributors`.

## Changes

- Hide and filter `@contributors` for profiles that are not a Wave creator or admin across suggestions, create, storm, draft, and edit paths.
- Treat directly typed unauthorized `@all` and `@contributors` strings as ordinary text without broadcast metadata or notifications.
- Preserve existing restricted mention metadata when a non-admin opens an unchanged existing message for editing.
- Keep the existing persisted `ALL` enum as the backward-compatible preference key, avoiding a migration or generated API change.
- Document that an existing non-admin `@contributors` token can be retained during an edit without rebroadcasting it.
- Update notification-setting coverage, user guides, the source Help Bot corpus, and the generated public corpus.

## Risk

- Level: Low
- Why: The UI narrows one mention path and relabels an existing preference without changing its stored representation.
- Rollback: Revert the feature commits; no data migration is involved.

## Review Notes

- Please focus on consistent sender gating and on the Broadcast mentions label accurately covering both admin-only broadcasts.
- [Backend PR #1949](https://github.com/6529-Collections/6529seize-backend/pull/1949) enforces sender authorization, preserves legacy editability, and applies the preference to contributor recipients.
- For the clearest transitional UX, deploy this frontend first and the backend immediately afterward.


prxt6529:
The cross-repository behavior behind the follow-up review is implemented by companion [backend PR #1949](https://github.com/6529-Collections/6529seize-backend/pull/1949):

- The backend rejects newly added `@all` and `@contributors` broadcasts from non-admins, so the docs describe the final coupled behavior rather than a frontend-only rejection.
- The backend intentionally interprets the persisted `ALL` preference as the shared `Broadcast mentions` key for both `@all` and `@contributors` recipients.
- Follow-up backend commit `7784bac9` preserves existing `@contributors` tokens on non-admin edits without rebroadcasting them; frontend commit `c855f754f` documents that edge case.

The pre-existing `Edit message...` localization debt is unrelated to this feature and was not changed.

prxt6529:
Follow-up review addressed in `4a4960885` after syncing with current `main`:

- User docs and the Help Bot corpus no longer say that directly typed unauthorized broadcast text is rejected.
- They now describe the actual create behavior: the string posts as ordinary text without broadcast metadata or notifications.
- Legacy edit wording now distinguishes retaining an existing `@contributors` broadcast token from adding a new broadcast mention.

Generated public Help Bot output was resynced, and documentation link validation passed.

PR #3824, opened 2026-08-25
Improve subscription and chat card alignment

ragnep:
## Issue

- Subscription status text in the profile header could truncate at desktop widths, while a fixed-width action area created excessive spacing for short labels.
- Embedded drop cards in chat rendered the artwork title lower than the media-type icon and rank badge.

## Fix

- Let the compact subscription row use its natural content width without truncating its status or reserving a fixed action width.
- Move the full-width subscription row below the profile identity until the `lg` breakpoint, where the longest supported status fits safely in the compact layout.
- Remove the drop-card heading's residual default top margin so the title, media-type icon, and rank badge share the same vertical center.

## Changes

- Adjusted the profile header subscription breakpoint and intrinsic-width layout.
- Kept narrow-mobile truncation behavior for genuinely constrained widths.
- Added a focused regression assertion for the chat drop heading margin.
- No wording, icons, colors, state logic, API behavior, or dependencies changed.

## Validation

- Visually checked subscription states including not set up, no upcoming selections, early warning, covered, unavailable, and loading.
- Checked responsive behavior at 320, 390, 840, 1023, 1024, and 1280 px widths with long and short action labels.
- Rendered the affected chat drop card with real API data and confirmed the icon, title, and rank badge have the same measured vertical center.
- `./bin/6529 run typecheck:changed` — passed.
- Focused `DropItemChat` Jest suite — passed.
- `./bin/6529 run react-doctor:diff` — 99/100; only the existing `UserPageHeaderClient` size advisory remains.
- `./bin/6529 run lint:changed` — attempted locally but exhausted the Node heap while the TypeScript project indexed unrelated nested worktrees; PR CI is the clean lint signal.

## Risk

- Level: Low
- Why: Tailwind-only layout adjustments plus one focused test assertion; no data, auth, or behavior changes.
- Rollback: Revert the single commit to restore the previous layout classes.

## Review Notes

- Focus on the `lg` subscription layout transition and the zero-margin heading in embedded chat drop cards.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile header responsiveness across different screen sizes.
  * Updated subscription status layouts to remain compact and prevent unwanted wrapping or overflow.
  * Adjusted drop title spacing for more consistent heading display.

* **Tests**
  * Strengthened coverage to verify drop titles render with the expected heading styling.

### DIFF
--- a/__tests__/components/waves/drops/DropItemChat.test.tsx
+++ b/__tests__/components/waves/drops/DropItemChat.test.tsx
@@ -60,7 +60,9 @@ describe("DropItemChat", () => {
     render(<DropItemChat href="https://base.com/p" dropId="d1" />);
     const links = screen.getAllByRole("link");
     expect(links[0]).toHaveAttribute("href", "/p");
-    expect(screen.getByText("Title")).toBeInTheDocument();
+    expect(screen.getByRole("heading", { name: "Title" })).toHaveClass(
+      "tw-m-0"
+    );
     expect(screen.getByTestId("position")).toHaveTextContent("2");
     expect(screen.getByTestId("media")).toHaveAttribute("media_url", "img");
     expect(screen.getByTestId("href-buttons")).toHaveTextContent("/p");
--- a/components/user/user-page-header/UserPageHeaderClient.tsx
+++ b/components/user/user-page-header/UserPageHeaderClient.tsx
@@ -314,7 +314,7 @@ export default function UserPageHeaderClient({
                     }`}
                   >
                     {canManageProfilePreferences ? (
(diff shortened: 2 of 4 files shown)


ragnep:
Reviewed the non-blocking advisories: the move to the `lg` compact breakpoint is intentional. The longest supported subscription state was verified at 1024px, while 840–1023px uses the full-width row below the identity header; narrow mobile retains constrained truncation. I am also keeping the focused `tw-m-0` assertion because it guards the exact residual heading-margin regression that caused the chat-card misalignment, while CI browser jobs cover the rendered integration. No follow-up code change is warranted.

PR #3826, opened 2026-08-25
Fix DM right sidebar deletion crash

GelatoGenesis:
## Issue

- Opening the right pane for a direct message owned by the logged-in user renders owner actions that call `useWaveDeleteFlow`.
- The messages desktop tree did not provide `WaveDeleteFlowProvider`, so React threw and the route error boundary displayed the Page of Doom.

## Fix

- Wrap the messages desktop content tree with the established `WaveDeleteFlowProvider`, matching the existing waves desktop provider boundary.

## Changes

- Add the deletion-flow provider to `MessagesDesktop`.
- Add a regression test that requests deletion from a consumer inside the messages tree and verifies the confirmation flow opens.
- No backend, API, route, copy, or visible styling changes.

## Validation

- `./bin/6529 run test __tests__/components/MessagesDesktop.test.tsx __tests__/components/MessagesLayout.test.tsx __tests__/components/waves/header/WaveHeaderOptions.test.tsx` — 3 suites and 8 tests passed.
- `./bin/6529 run typecheck` — passed.
- `./bin/6529 run lint:changed` — passed with zero warnings.
- `./bin/6529 run react-doctor:diff` — 100/100, no findings.

## Risk

- Level: Low
- Why: additive context boundary around the existing messages desktop tree; no data, API, authentication, or navigation contract changes.
- Rollback: revert the provider wrapper and regression test.

## Review Notes

- Confirm provider placement remains parallel with `WavesDesktop` and covers right-sidebar portals through React context propagation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved wave deletion from the desktop messages view.
  - Selecting a wave for deletion now opens the correct confirmation modal.
  - Added coverage to help ensure the deletion flow remains reliable.

### DIFF
--- a/__tests__/components/MessagesDesktop.test.tsx
+++ b/__tests__/components/MessagesDesktop.test.tsx
@@ -1,14 +1,10 @@
 import MessagesDesktopWithProvider from "@/components/messages/MessagesDesktop";
-import { useWaveDeleteFlow } from "@/components/waves/header/options/delete/WaveDeleteFlowContext";
+import WaveHeaderOptions from "@/components/waves/header/options/WaveHeaderOptions";
 import type { ApiWave } from "@/generated/models/ApiWave";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import type { ReactNode } from "react";
-
-const mockWaveDeleteModal = jest.fn(
-  ({ isOpen }: { readonly isOpen: boolean }) =>
-    isOpen ? <div data-testid="delete-wave-modal" /> : null
-);
 
 jest.mock("@/components/brain/ContentTabContext", () => ({
   ContentTabProvider: ({ children }: { readonly children: ReactNode }) => (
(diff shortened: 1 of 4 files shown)


GelatoGenesis:
Addressed the clarity suggestion in `7b8a0a85c` by naming the default import `MessagesDesktopWithProvider`. I am leaving out the negative test because the inner unwrapped component is intentionally private; exporting it only to assert an unsupported render path would widen the component API and couple the regression test to implementation details. The positive test fails if the exported provider boundary is removed and directly protects the reported crash.

GelatoGenesis:
Reviewed the remaining 6529bot lanes on the latest head:

- **GLM advisory:** repository search confirms there is no ancestor `WaveDeleteFlowProvider` on the exported `MessagesDesktop` path. The other providers belong to sibling `/waves` desktop and native About branches, so this wrapper does not nest or shadow another instance. `WaveDeleteFlowProvider` and `WaveDeleteModal` dependencies were inspected, and the existing real `WaveHeaderOptions` and `WaveDeleteModal` suites pass alongside the new provider-boundary regression test.
- **Responsiveness runner:** deferring this as runner infrastructure rather than changing this PR. The report is bound to `32a02a4f5` and shows 101–124 local `_next/static` requests aborted on every native-simulation route before hydration; the same run passes web desktop, responsive web, and Electron. This PR changes only a React provider boundary and its test—no Next config, asset serving, environment, Capacitor, or app-shell code. On latest head `7b8a0a85c`, the required App PR CI Playwright smoke, quality/contracts production build, installed-app checks, CodeQL, SonarCloud, Snyk, and all other check-rollup entries are successful. Fixing the external native-simulation asset server belongs in the review runner and would be unrelated scope here.

GelatoGenesis:
Addressed the valid GLM test-coverage findings in `9fca29bd8`:

- The regression test now mounts the real `WaveHeaderOptions` owner menu beneath the exported `/messages` provider boundary.
- It opens the real Delete action and real `WaveDeleteModal`, cancels, and reopens it, covering the reusable context lifecycle that previously crashed.
- The focused `MessagesDesktop`, `WaveHeaderOptions`, and `WaveDeleteModal` suites pass; changed-file typecheck/quality passes; React Doctor is 100/100.
- Provider audit remains clean: `/messages` has no ancestor `WaveDeleteFlowProvider`; the other instances are sibling `/waves` desktop and native About paths, so there is no nested provider or shadowed modal state.

I also traced the responsiveness finding to its precise cause. The workflow already sets both `ASSETS_FROM_S3=false` variables. The newly merged base branch intentionally runs `NATIVE_IOS_BOOTSTRAP_SCRIPT` before hydration; in a fresh iOS context it persists `is-native-ios=true` and calls `location.reload()`. The runner injects an iOS Capacitor shim into every fresh native page, records the first navigation’s expected `_next/static` aborts, runs routes in parallel, and only prewarms `web-desktop`/`web-mobile`. That exactly explains why all native-only requests show `net::ERR_ABORTED` while every other context succeeds. The safe correction is runner-side: prewarm/seed the native iOS marker (or disregard aborts from the intentional bootstrap navigation), then assess the settled load. Changing this DM PR’s asset config or removing the production EULA bootstrap reload would be unrelated and would reintroduce the hydration defect that bootstrap protects.

/6529bot review followup glm-swarm responsiveness

GelatoGenesis:
Addressed the latest valid GLM items in `cbb291aa7`:

- The integration test now uses the shared complete `createMockApiWave` fixture and a full App Router stub.
- After real owner-menu open → cancel → reopen, it now authenticates and confirms deletion, asserting the `waves/dm-wave` DELETE, drop invalidation, and `/waves` navigation.
- Mobile/provider audit: `WaveHeaderOptions` has one production render path through `WaveHeader` → `WaveContent`. Desktop `/waves` is covered by `WavesDesktop`, desktop `/messages` by this PR, and mobile `/waves` plus `/messages` About views by the existing `BrainMobileAbout` provider. Adding another mobile wrapper would create the duplicate-provider condition the review warns against.
- Provider ordering audit: `WaveDeleteFlowProvider` itself depends only on its modal state and viewport hook. The modal’s auth, React Query, and router dependencies come from root providers above `ContentTabProvider`; it has no dependency on `ContentTabProvider`, so the current ordering is valid.

The responsiveness rerun remains a confirmed harness mismatch. The app’s own native Playwright setup in `tests/support/surfaceSimulation.ts` calls `seedIosEulaConsent` before navigation and seeds both `NATIVE_IOS_COOKIE` and `CONSENT_EULA_COOKIE`. The external runner injects the iOS Capacitor shim but seeds neither, so `NATIVE_IOS_BOOTSTRAP_SCRIPT` correctly performs its one-time reload and the runner misclassifies the resulting first-navigation aborts. Its own environment already has `ASSETS_FROM_S3=false`. The runner should mirror the repository harness by seeding those cookies before `page.goto`; changing the DM provider or production asset config cannot fix that simulation contract.

/6529bot review glm-swarm

GelatoGenesis:
Resolved the final valid product findings in `91a47b8c2`:

- `WaveDeleteModal` now derives its return route from the deleted wave. DMs return to `/messages`; ordinary waves retain `/waves`.
- Successful deletion invalidates both drop caches and the full wave/sidebar query set before navigation (`onWaveCreated` is the existing shared all-wave invalidator, already used after wave edits as well as creation).
- The DM integration test now asserts one DELETE, both invalidations, their ordering before navigation, and the `/messages` destination.
- Modal tests now cover API rejection and cancelled authentication: the dialog stays usable, controls recover, no navigation occurs, and caches remain untouched.

The partial context in the focused integration test remains intentional: it exposes exactly the two context calls consumed by the real modal and asserts both. A full dozens-of-members mock would obscure the regression without increasing runtime coverage. The modal also retains its existing local `mutating` guard plus disabled/loading button behavior against double submission.

Local validation: all three deletion-related Jest suites pass; changed-file check/typecheck/format/lint pass; full Jest + Playwright typecheck ratchets pass; React Doctor 100/100.

PR #3827, opened 2026-08-26
Keep hidden actions visible on hybrid touch laptops

ragnep:
## Issue

- Hybrid Windows touch laptops can keep a desktop layout while reporting that reliable CSS hover is unavailable.
- Controls revealed only by capability-gated hover then remain invisible even though the trackpad still moves the cursor. Reported examples include Wave pinning and profile social-statement deletion.

## Fix

- Use the existing touch-only variant, which is the exact complement of Tailwind's reliable-hover query, to keep affected actions visible when CSS hover cannot fire.
- Preserve the existing hover-reveal design on ordinary mouse desktops and the existing responsive/mobile layouts on phones.
- Strengthen the repository contract so future hover-hidden controls must provide an effective no-hover fallback, including across responsive cascade gates.

## Changes

- Add no-hover fallbacks for Wave pin/unpin, profile statement deletion, activity-log copy, profile About/banner/name/classification editing, link-card actions, leaderboard actions, unread dismissal, and gallery author links.
- Preserve keyboard focus reveals and add missing accessible button semantics where these controls were touched.
- Replace the prior visibility baseline with an AST-based audit covering display, visibility, pointer events, opacity, collapsed dimensions, self-hover, group-hover, partial opacity, responsive gates, and documented literal-level exemptions.

## Validation

- 80 focused component and contract tests passed before the final cascade hardening; the final contract suite passes 11/11.
- Changed-file lint passed.
- Changed-file typecheck passed for 593 TypeScript files.
- Full changed-file quality check passed and confirmed the branch was current with origin/main.
- React Doctor scored 99/100; its two warnings are pre-existing structural observations unrelated to this fix.
- Browser QA against the staging API confirmed desktop mouse reveal, keyboard focus reveal, a successful pin/unpin round trip with state restored, and the 390px phone layout with no horizontal overflow and its existing mobile Edit profile path intact.
- A real Surface device was not available. Exact no-hover coverage is provided by the compiled CSS complement and responsive-cascade contract rather than claimed as real-device evidence.

## Risk

- Level: Low
- Why: CSS visibility and focus behavior only; no API, auth, data-model, dependency, route, or deployment changes. Desktop hover and phone layout behavior remain unchanged.
- Rollback: Revert the single commit to restore the previous per-control visibility classes and contract.

## Review Notes

- Please focus on the touch-only fallback coverage, the documented exemption set, and the responsive opacity override used by the gallery author link.
- The global touch-device classifier was intentionally not changed because hybrids must retain the desktop layout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved touch-device usability by keeping key actions, edit controls, pin controls, copy buttons, and overlays visible and interactive without hover.
  * Preserved desktop hover and keyboard-focus behavior while adding clearer focus styling.
  * Improved responsive visibility for wave gallery authors, leaderboard actions, and profile controls.
  * Prevented unintended form submissions from action buttons by explicitly defining their button behavior.

* **Tests**
  * Expanded coverage for responsive touch, hover, focus, visibility, and interaction states.

### DIFF
--- a/__tests__/components/ChatItemHrefButtons.test.tsx
+++ b/__tests__/components/ChatItemHrefButtons.test.tsx
@@ -463,20 +463,23 @@ describe("ChatItemHrefButtons", () => {
     expect(screen.getByRole("button", { name: "Link actions" })).toBeVisible();
   });
 
-  it("keeps the overlay trigger visible when the CSS hover query is unavailable", () => {
+  it("keeps the overlay trigger operable when the CSS hover query is unavailable", async () => {
+    const user = userEvent.setup();
     render(
       <div className="tw-group/link-card tw-relative">
         <ChatItemHrefButtons href="https://a" layout="overlay" />
       </div>
     );
 
-    expect(
-      screen.getByRole("button", { name: "Link actions" }).parentElement
-    ).toHaveClass(
+    const trigger = screen.getByRole("button", { name: "Link actions" });
+    expect(trigger.parentElement).toHaveClass(
       "tw-pointer-events-none",
       "tw-opacity-0",
       "touch-only:tw-pointer-events-auto",
       "touch-only:tw-opacity-100"
     );
+
+    await user.click(trigger);
+    expect(screen.getByRole("button", { name: "Copy link" })).toBeVisible();
   });
 });
--- a/__tests__/components/ChatItemHrefButtons.test.tsx
+++ b/__tests__/components/ChatItemHrefButtons.test.tsx
@@ -462,4 +462,21 @@ describe("ChatItemHrefButtons", () => {
 
     expect(screen.getByRole("button", { name: "Link actions" })).toBeVisible();
   });
+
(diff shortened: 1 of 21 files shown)


ragnep:
Review follow-up on exact head c2aea96f27937d5826e313b0a4ce896aa48f9f61:

- Resolved the remaining hybrid-hover finding by giving the primary unread control a touch-only full-opacity fallback and adding an interaction regression test for both unread actions.
- Hardened audited contract exemptions against substring matches and made the podium mappings one-to-one.
- The repository follow-up reviewer reports no new findings; required PR checks, production build, both Playwright lanes, CodeQL, SonarCloud, Snyk, DCO, and CodeRabbit are green.
- The separate responsiveness advisory failed only in simulated Capacitor because its runner aborted local _next/static assets. Its web desktop, web mobile, and Electron matrices were clean, this PR does not change asset serving or Capacitor code, and the repository production-build and browser lanes pass. I am treating that as unrelated runner infrastructure rather than broadening this UI fix into deployment/build changes.

PR #3830, opened 2026-08-26
Replace unhealthy Beanstalk instances from ELB health

prxt6529:
## Issue

- The production Elastic Beanstalk Auto Scaling group used EC2-only health checks, so instances whose application stopped responding were removed from ALB traffic but were not automatically replaced.

## Fix

- Configure the Beanstalk-managed Auto Scaling group to use ELB health checks while retaining the existing homepage (`/`) target health check.
- Give newly launched instances a 300-second grace period before ELB health can trigger replacement.

## Changes

- Add `.ebextensions/autoscaling-health.config` with the AWS-documented `AWSEBAutoScalingGroup` resource override.

## Validation

- Parsed the configuration as YAML and asserted `HealthCheckType: ELB` and `HealthCheckGracePeriod: 300`.
- Repository-wide lint completed successfully.
- The local optimized Next.js production compilation was started with CI's public production endpoint values but remained nonterminal on the local runner and was stopped; PR CI remains the authoritative production build.

## Risk

- Level: Medium
- Why: This changes production instance-replacement behavior. Persistent homepage health-check failures will now cause Auto Scaling to replace affected instances.
- Rollback: Remove the `.ebextensions` override and deploy a subsequent application version to restore EC2-only health checks.

## Review Notes

- Confirm that the 300-second startup grace period is appropriate for current Beanstalk boot time and that `/` remains the intended ALB health-check path.

## Summary by CodeRabbit

* **New Features**
  * Improved deployment reliability by configuring automatic instance health checks through the load balancer.
  * Added a 300-second grace period to allow new instances to initialize before health evaluation.

### DIFF
--- a/.ebextensions/autoscaling-health.config
+++ b/.ebextensions/autoscaling-health.config
@@ -1,0 +1,6 @@
+Resources:
+  AWSEBAutoScalingGroup:
+    Type: AWS::AutoScaling::AutoScalingGroup
+    Properties:
+      HealthCheckType: ELB
+      HealthCheckGracePeriod: 300


prxt6529:
Verified this finding against the current AWS Elastic Beanstalk documentation. AWS explicitly prescribes this exact `Resources -> AWSEBAutoScalingGroup -> AWS::AutoScaling::AutoScalingGroup` override with `HealthCheckType: ELB` and `HealthCheckGracePeriod: 300` for application-level replacement: [Auto Scaling health check setting for your Elastic Beanstalk environment](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environmentconfig-autoscaling-healthchecktype.html).

The suggested `aws:autoscaling:healthcheck` option namespace is not listed among Elastic Beanstalk's supported general configuration namespaces. Operationally, the environment's existing target check uses `/`, and both replacement instances registered healthy against it during incident recovery. The 300-second grace period also matches AWS's documented example. Retaining the configuration unchanged; an environment deployment is outside this PR's scope.

prxt6529:
Bounded follow-up on `b8e39166e470`:

- The target environment is a `LoadBalanced` standard `WebServer` tier with an application load balancer, so the single-instance/worker-tier advisory does not apply.
- Application deployments use `RollingWithAdditionalBatch`; the environment has min capacity 2 and max capacity 5, so deployment capacity protection is already configured independently of this health-check override.
- The existing ALB process check is HTTP `/` on port 3001 (15-second interval, 5-second timeout, 3 healthy / 5 unhealthy threshold, matcher 200). Both incident-replacement instances registered healthy against it, and the environment is currently Green.
- No other `.ebextensions` or workflow configuration sets `HealthCheckType` or `HealthCheckGracePeriod`.
- The branch is current with `main`, with no review threads, no Sonar issues, and no failing task-owned checks in the captured snapshot.

The 300-second grace period and resource override remain the exact pattern documented by AWS. No source change is needed from this follow-up.

PR #3861, opened 2026-09-01
Serialize wallet session refreshes across browser tabs

prxt6529:
## Issue

- Multiple browser tabs could refresh the same rotating wallet-auth session concurrently. Both requests could send the same cookie, allowing response ordering to leave the browser with a credential the server had already replaced.

## Fix

- Serialize web session refreshes across browser tabs with an address-scoped Web Locks API lock while retaining the existing same-tab in-flight deduplication.

## Changes

- Added an address-normalized, web-only cross-tab refresh lock using the same key as in-process coordination.
- Preserved the existing native/desktop refresh path and added safe fallback when Web Locks are unavailable or acquisition fails before the refresh begins.
- Preserved abort propagation and prevented fallback from retrying a refresh task that already started.
- Added focused coverage for lock naming, acquisition failure, task failure without retry, and abort while queued.
- No UI, generated API client, cookie format, persistence, or authentication-signing changes.

## Validation

- Focused session-v2 utility suite: 41 tests passed.
- Changed-file lint passed.
- Changed-file TypeScript check passed.
- Debt ratchet passed with no oversized source files.

## Risk

- Level: Medium
- Why: The change affects browser authentication refresh scheduling but not credential contents or signing. Unsupported browsers retain the existing behavior and are protected by the companion backend fix.
- Rollback: Revert this PR and redeploy the web frontend.

## Review Notes

- Review the lock scope, abort-signal propagation, and no-Web-Locks fallback.
- Companion backend: https://github.com/6529-Collections/6529seize-backend/pull/1959
- Deploy after the companion backend API change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session refresh reliability across web, native, and desktop clients.
  * Web sessions now continue refreshing when cross-tab coordination is unavailable or fails.
  * Improved handling of refresh cancellations and errors, preventing unnecessary retries.
  * Added recovery behavior so later refresh attempts can proceed normally after a failed coordination attempt.
  * Improved consistency when coordinating refresh activity across browser tabs.


prxt6529:
Addressed the auth-review findings in a1d16b651: extracted refresh coordination so the original service stays below the debt threshold; added a lock-acquisition fallback that runs only when the refresh task never started and the error is not an AbortError; preserved lock-level abort propagation through the existing abort telemetry wrapper; and added coverage for acquisition failure, no retry after task start, and abort while queued followed by a successful retry. The focused auth suite, changed-file lint/typecheck, and debt ratchet now pass.

PR #3868, opened 2026-09-01
Fix mobile drop owner action dialogs

ragnep:
## Issue

- On narrow/touch layouts, an owner could tap enabled **Resubmit Drop** or **Delete Drop** actions but see no response.
- The actions did fire: their dialogs rendered underneath the mobile drop-detail panel because the parent and nested overlays used conflicting stacking levels.
- The resubmission form also used excessive introductory spacing and mismatched footer button sizing on small screens.

## Fix

- Render both owner-action overlays with Headless UI dialog primitives at the established nested-dialog layer (`z-1020`).
- Keep the mobile detail panel unchanged so the correction stays scoped to dialogs that must appear above it.
- Preserve existing submission/deletion behavior while adding focus management, Escape/outside close behavior, accessible names, and visible loading/error-safe interaction states.

## Changes

- Migrate the meme resubmission and drop deletion overlays to accessible dialog/backdrop/panel/title primitives.
- Add focused tests for stacking, dialog semantics, dismissal, autofocus, success, mutation rejection, and submission-limit guidance.
- Give the narrow-screen modal title a dedicated wrapping column beside the close action, with balanced supporting copy.
- Tighten the resubmission acknowledgement hierarchy and vertical rhythm.
- Reflow small-screen form actions into equal-size secondary controls above a full-width primary action; keep the existing desktop row.

## Validation

- Focused Jest suites for the resubmission modal/action, deletion modal, final form step, and preview screen.
- `./bin/6529 run react-doctor:diff` (99/100; two existing motion import advisories, no new correctness findings).
- Pixel 5 touch-size browser reproduction: both dialogs are visible above the detail panel and support keyboard dismissal/focus behavior.
- Pixel 5 header geometry: the title wraps to two lines, the title and close control have zero overlap, and supporting copy uses balanced wrapping.
- Responsive browser checks at mobile and desktop sizes: footer controls are consistent and non-cluttered.
- Authenticated staging fixture created and a replacement submission persisted for the real competition workflow. The in-app browser's agent session has no attached wallet signer, so the final signed UI request was also validated through the app's authenticated submission modules.

## Risk

- Level: Low
- Why: changes are limited to two owner-action dialogs, their focused tests, and resubmission layout classes; API payloads and ownership rules are unchanged.
- Rollback: revert the two commits to restore the previous overlays and layout.

## Review Notes

- Please focus on nested-dialog layering and focus restoration from the mobile drop-detail panel.
- The reported account was below the active Main Stage submission limit; the apparent no-op was a visual stacking defect, not a disabled eligibility path.

## Summary by CodeRabbit

- **New Features**
  - Delete failures now restore the action button, allowing users to retry.
  - Added clearer messaging when resubmission limits are reached.
  - Improved semantic modal titles, accessibility, and keyboard interactions.

- **Style**
  - Refined delete and submission modal spacing, typography, borders, shadows, and responsive layouts.
  - Improved wrapping for long titles and descriptions.
  - Buttons now adapt more consistently across screen sizes.

- **Tests**
  - Expanded coverage for modal behavior, accessibility, styling, deletion failures, and resubmission limits.


ragnep:
Addressed the valid previous-head review findings in `d4e1cf1257`:

- restored the exact original trash-icon path data;
- moved the hidden `DialogTitle` inside `DialogPanel` and added an explicit visible `data-autofocus` close target;
- removed the duplicate wrapper `role="dialog"`;
- kept the transition mounted with `show={isOpen}` so leave classes can run, while preserving the server-side `document` guard;
- made deletion success/failure assertions await the asynchronous callbacks and verify recovery;
- removed the acknowledgement list's default top margin and added focused regression coverage.

The CodeRabbit docstring-coverage warning is intentionally deferred: these are small React UI components and callbacks in a codebase that does not use function docstrings for this pattern; adding them would be noise rather than useful API documentation.

The generic responsiveness workflow produced no routes, so responsive evidence comes from the authenticated local flow: Pixel 5 touch geometry showed the wrapped title and close action have zero overlap, all footer buttons are 40px high, and both nested dialogs render at the expected layer.

PR #3884, opened 2026-09-03
Finish Release Bus retirement and restore separate automatic E2E

GelatoGenesis:
## Issue

Finish retiring the old deployment runtime and skill prompts while preserving the original deployment behavior: Web Deploy finishes before automatically triggered E2E, and autonomous release notes remain enabled.

## Fix

- Restore separate staging and production E2E dispatch workflows after successful Web Deploy completion, preserving the original deploy run ID for CI-wave replies.
- Remove the staging runtime directory alias after migrating the live host to `.deploy`, preserving its lock inode, current release, rollback releases, and PM2 configuration.
- Update the deployment skill launch prompt and guidance. Suppressing release notes now requires an explicit user request; internal maintenance keeps them enabled.

## Validation

- 38 deployment, E2E, notification, and staging-runtime tests passed.
- Both dispatch scripts passed isolated execution checks for a new run, duplicate reuse, another repository, and a failed deployment.
- The live staging host is healthy on the same application version with PM2 using `.deploy` paths.
- Changed-file lint and formatting passed; Jest typecheck ratchet checked separately.

## Risk

Medium: workflow orchestration changes. Builds, artifact verification, health checks, and E2E packs remain intact. E2E reports its own result after deployment; it no longer keeps Web Deploy open. Ordinary shared-branch history is preserved.

## Review Notes

- The dispatcher deliberately runs the trusted verifier/notifier from `main`, then checks out the deployed application's exact source for E2E. This preserves the former separate-dispatch trust model for staging too.
- GitHub's actual workflow-run API returns the dynamic `run-name` in both `name` and `display_title`; the exact run-title filter is intentional. Existing runs are reused regardless of conclusion; rerunning their failed jobs preserves the deployment correlation.
- Manual E2E reruns now explicitly require the successful deployment run ID, matching the validation already enforced by the workflow. Automatic dispatch supplies it. Ordinary deployments need no commit or run-ID inputs.
- The dispatch lookup allows three minutes and distinguishes a failed POST while checking for an accepted request before any retry. The resolver tests independently check run-level and job-level success.
- The live staging host has already migrated to the physical `.deploy` directory with neutral PM2 paths. The old directory is removed, the lock inode is preserved, and health/version checks passed.
- The deployment skill already keeps release notes enabled for internal maintenance; only an explicit user request enables opt-out and omits grouping/publication metadata.

### DIFF
--- a/.github/workflows/build-upload-deploy-prod.yml
+++ b/.github/workflows/build-upload-deploy-prod.yml
@@ -331,20 +331,6 @@ jobs:
           if-no-files-found: ignore
           retention-days: 90
 
-  automatic-production-e2e:
-    name: Automatic production E2E
-    needs:
-      - build-upload-deploy
-      - notify-production-deployment
-    uses: ./.github/workflows/production-e2e.yml
-    with:
-      trusted_deployed_sha: ${{ github.sha }}
-      canonical_deploy_call: true
-      parent_deploy_run_id: ${{ github.run_id }}
-    secrets:
-      CI_PIPELINES_ALERT_API_AUTH: ${{ secrets.CI_PIPELINES_ALERT_API_AUTH }}
-      CI_PIPELINES_ALERT_SECRET: ${{ secrets.CI_PIPELINES_ALERT_SECRET }}
-
   notify-production-deployment:
     name: Notify production deployment outcome
     if: always()
--- a/.github/workflows/build-upload-deploy-prod.yml
+++ b/.github/workflows/build-upload-deploy-prod.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       release_note_opt_out:
-        description: Skip autonomous release notes for an internal operation
+        description: Suppress autonomous release notes only when explicitly requested by the user
         type: boolean
         required: false
         default: false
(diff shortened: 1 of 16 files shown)


GelatoGenesis:
Checked the remaining findings against GitHub and the executed scripts:

- The name-filter finding does not match this repository's REST responses. `GET /repos/6529-Collections/6529seize-frontend/actions/runs/33756559344` returns both `name` and `display_title` as `Staging E2E [automatic]`, matching its dynamic `run-name`. The exact-title filter is intentional and the isolated new-run/reuse cases pass for both dispatchers.
- `split("deploy_run_attempt=")` produces exactly two segments in each resolver. The URL uses `${deploy_run_attempt}`, without an equals sign, so it is not another separator. Executing the split confirmed both the run segment and job segment independently contain their success checks; all 38 focused tests pass.
- Manual reruns now explicitly require the successful deploy run ID, and the docs explain it. Automatic dispatch fills it in. Staging deliberately uses the verifier/notifier from main, then checks out the deployed source for tests.
- The manual production form's opt-out description is now corrected as well as the skill. Notes stay enabled unless the user explicitly requests suppression.
- The live staging runtime has already migrated to physical `.deploy` and neutral PM2 paths, with its lock inode, rollback data and healthy application version verified.

No further code change is needed for the two disproven follow-up findings. The dispatcher allows three minutes for visibility and reports a failed POST separately while avoiding a duplicate dispatch after an uncertain response.

PR #3922, opened 2026-09-09
fix(cms): resolve published pages with display-case handles

punk6529:
## Issue
Published custom profile packages can retain display-case handles, while requests use the canonical lowercase handle. A valid page such as `/Punk6529/Gallery/index.html` can consequently return not found.

## Fix
Normalize only the profile-handle segment when matching routes and producing internal links. Preserve case-sensitive page names, query values, and fragments.

## Changes
- Apply the same handle normalization to aliases, safe redirects, page links, and nested navigation.
- Preserve alias-cycle detection, unsafe-redirect rejection, external URLs, and the original package objects; reject conflicting canonical routes independently of manifest order.
- Add regression tests and update profile troubleshooting and the Help Bot corpus.

## Validation
- All four CMS runtime suites passed: 32 tests, including 17 route tests.
- Help index synchronization and documentation link validation passed (299 files).
- Whitespace check passed.
- Changed-file type check passed; React Doctor found no issues in the two changed source files.
- CI passed changed-file lint, quality/contracts, production build, Playwright smoke and critical shell, CodeQL, SonarCloud, Snyk, and DCO.

## Risk
Low: scoped to custom profile route matching and generated links. No package format, backend, or data migration changes. Revert this change to roll back.

## Review Notes
Check that only the handle is normalized, including handle-only links with queries or fragments, and that aliases retain cycle protection.

### DIFF
--- a/__tests__/lib/profile-cms/runtime/routes.test.ts
+++ b/__tests__/lib/profile-cms/runtime/routes.test.ts
@@ -46,6 +46,25 @@ describe("profile CMS runtime routes", () => {
     }
   );
 
+  it.each([false, true])(
+    "rejects canonical route collisions regardless of order (reversed: %s)",
+    (reversed) => {
+      const routes: CmsPackageV1["payload"]["routes"] = [
+        { path: "/Punk6529/index.html", kind: "page", page_id: "page-home" },
+        {
+          path: "/punk6529/index.html",
+          kind: "redirect",
+          target: "/punk6529/Other/index.html",
+        },
+      ];
+      const cmsPackage = withRoutes(reversed ? routes.toReversed() : routes);
+      expect(resolveCmsRoute(cmsPackage, "/punk6529/index.html")).toEqual({
+        kind: "not_found",
+        reason: "route_missing",
+      });
+    }
+  );
+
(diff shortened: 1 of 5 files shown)


punk6529:
Addressed the canonical-route collision: resolution now requires exactly one normalized match and fails closed for conflicting entries. Both manifest orderings have regression coverage; all 17 focused route tests pass.

Deferring a route index/cache until there is a demonstrated performance need; this patch keeps resolution scoped and avoids changing package lifetime or cache behavior. Lowercasing the hexadecimal letters in a percent-encoded octet does not change its decoded value, and this function does not decode or re-encode the segment. The safe-relative-URI guard remains in place.

punk6529:
The Help Bot corpus was validated through `help-index:sync`, which checks required record fields, unique IDs, route references, and source references before generating the public index. All 216 published records passed. The lookup implementation is unchanged.

The advisory responsiveness run produced no runner summary and is not a passing test. This PR changes route resolution and links without changing rendering markup; live staging and production browser checks will exercise the published pages during the authorized rollout. The reverse-link lookup note is deferred: ambiguous packages are rejected when resolving the destination, and reverse links do not authorize access.

PR #547, opened 2024-10-04
Wave chat mobile fixes

analyticsflowee:

PR #3951, opened 2026-09-11
Fix pinned waves after expired sessions

GelatoGenesis:
## Issue

An expired saved session can receive the public wave overview in response to a pinned-waves request. The sidebar then labels every returned wave as pinned, and unpinning briefly removes a wave before the failed request restores it. Report: https://6529.io/waves/e933d8d6-0c78-4e8f-aa12-f67d8c11b4dc?serialNo=1419915.

## Fix

Only render server-confirmed pins and authenticate before starting a pin or unpin mutation. Await sidebar mutations so asynchronous failures reach the existing error toast.

## Changes

- Filter fetched and cached pinned-wave results by the server's `pinned` flag; stop pagination when a response includes unconfirmed pins.
- Recheck the active wallet and JWT role after authentication and before optimistic cache writes and immediately before the API request to prevent an account switch from applying the action to another profile. Accept the owner profile ID used by normal backend logins, or a legacy absent role; reject other profile roles.
- Add regression coverage for public responses, contaminated cache, rejected authentication, account/proxy changes, and asynchronous failures.
- Move pin-control labels, accessible names, validation, and errors into the existing message dictionary; document English fallback and verify all five supported locales.
- Update pinning and session-recovery help, including generated public help data.

## Validation

- 123 focused pin-control, wave-list, and message tests passed after review fixes; 150 existing posting/auth/session-recovery tests also passed.
- Latest GitHub CI passed the production build, both browser packs, and all related Jest batches. Changed-file lint/typecheck and formatting also passed locally.
- React Doctor: 100/100. Help index sync and 304 documentation link checks passed.
- Staging deployment [34596449445](https://github.com/6529-Collections/6529seize-frontend/actions/runs/34596449445) passed artifact, live-version, and health checks. All 16 API cases passed; desktop and 390×844 browser pin/unpin survived reloads, both UI test posts were confirmed through the API, and original pins were restored. Physical iPhone testing has not been performed; the native shell checkout is unavailable locally.

## Risk

- Level: Medium
- Why: changes shared pinning behavior and adds authentication before mutations; failed requests retain the existing rollback behavior.
- Rollback: revert this PR through the normal deployment workflow. No API shape or database changes.

## Review Notes

Deploy backend https://github.com/6529-Collections/6529seize-backend/pull/2002 (`api` only) before this frontend in each environment. The frontend filter also tolerates the old backend during rollout. Existing native reconnect/posting recovery from #3950 is retained and covered by the focused tests.

### DIFF
--- a/__tests__/components/brain/left-sidebar/waves/BrainLeftSidebarWavePin.test.tsx
+++ b/__tests__/components/brain/left-sidebar/waves/BrainLeftSidebarWavePin.test.tsx
@@ -1,11 +1,10 @@
-import { render, screen } from "@testing-library/react";
+import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import BrainLeftSidebarWavePin from "@/components/brain/left-sidebar/waves/BrainLeftSidebarWavePin";
 import {
   MAX_PINNED_WAVES,
   usePinnedWavesServer,
 } from "@/hooks/usePinnedWavesServer";
-import { useMyStream } from "@/contexts/wave/MyStreamContext";
 import { useAuth } from "@/components/auth/Auth";
 
 // Mock ResizeObserver
@@ -27,7 +26,6 @@ jest.mock("react-tooltip", () => ({
 jest.mock("@fortawesome/react-fontawesome", () => ({
   FontAwesomeIcon: () => <svg data-testid="icon" />,
 }));
-jest.mock("@/contexts/wave/MyStreamContext");
 jest.mock("@/hooks/usePinnedWavesServer");
 jest.mock("@/components/auth/Auth");
 
@@ -56,7 +54,6 @@ const proxyAuth: AuthMock = {
   connectedProfile: { handle: "testuser" },
   activeProfileProxy: { id: "proxy-1" },
 };
-const mockedUseMyStream = useMyStream as jest.Mock;
 const mockedUsePinnedWavesServer = usePinnedWavesServer as jest.Mock;
 const mockedUseAuth = useAuth as jest.Mock;
 
@@ -67,10 +64,9 @@ function setup(
   auth: AuthMock = connectedAuth,
   compact = false
 ) {
-  mockedUseMyStream.mockReturnValue({
-    waves: { addPinnedWave, removePinnedWave },
-  });
   mockedUsePinnedWavesServer.mockReturnValue({
+    pinWave: addPinnedWave,
+    unpinWave: removePinnedWave,
     pinnedIds: storedPinned,
     isOperationInProgress: jest.fn().mockReturnValue(false),
     canPinWave: jest.fn().mockImplementation(canPinWave),
@@ -84,6 +80,8 @@ function setup(
 describe("BrainLeftSidebarWavePin", () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    addPinnedWave.mockReset().mockResolvedValue(undefined);
+    removePinnedWave.mockReset().mockResolvedValue(undefined);
     Object.defineProperty(navigator, "maxTouchPoints", {
       configurable: true,
       value: 0,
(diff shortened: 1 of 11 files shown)


GelatoGenesis:
Addressed in fc25afbb88fca3bdac4ae4f1284a42097d70608d:

- Extracted the sidebar pin/unpin copy, accessible names, validation, and errors into `waves.sidebar.pinControl.*` messages. The component follows browser locale and formats the limit with the existing helper. Documented non-source-locale English fallback in the sidebar workstream tracker and tested all five supported locales.
- A viewer change after successful authentication now throws so the caller shows its error toast. Recheck the viewer after asynchronous query cancellation and before optimistic writes, as well as immediately before the API request.
- Added coverage for rejected authentication, pending-operation cleanup, and no cache writes after rejected auth or a viewer switch. Documented why pagination stops if the server ignores the pinned-only filter.

The advisory suggestion to support in-place React Query cache mutation is deferred: these cache updates are immutable, and mutating data in place violates the cache contract. Recomputing on a derived key would not fix missing observer notifications. Authentication returning `success:false` intentionally preserves the auth flow's own feedback, skips mutations and cache writes, and does not produce a second pin error toast.

114 focused tests, changed-file lint/typecheck, docs checks, and React Doctor (100/100) pass. The responsiveness bot did not execute tests because its external workflow called the removed `6529 install:frozen` command; deployed staging will receive direct desktop/mobile-viewport smoke coverage.

GelatoGenesis:
Targeted staging smoke found the new pin guard rejecting the QA primary profile before any mutation, while posting from that same profile succeeded and persisted. The guard used saved wallet-role metadata, but AuthProvider derives the effective role from the current JWT; a valid local JWT does not synchronize stale saved role metadata.

The follow-up uses the effective JWT role and still rejects a changed wallet, proxy JWT, missing JWT, or malformed JWT before optimistic writes and API calls. Added primary-JWT/stale-saved-role regression coverage and retained account/proxy-switch checks. All 119 focused tests, changed-file TypeScript/lint, and React Doctor pass. Production remains pending successful repeat staging smoke after review and CI.

GelatoGenesis:
The repeat staging pin test still rejected the primary profile, so production remains on hold. The saved-role correction was insufficient. The concrete contract is in backend `auth.routes.ts` → `resolveAuthenticatedRole`: a normal login with no requested proxy sets the JWT role to the signing profile ID. AuthProvider only treats that role as a proxy if it matches a received proxy, so a truthy role alone does not identify a proxy.

The guard now accepts the current primary profile ID (or a legacy absent role), while retaining wallet/viewer checks and rejecting all other profile roles. The test fixture now uses a real primary profile ID and the backend's primary-role token shape. Both pin and unpin tests failed against the previous guard before the fix, and pass with this correction. Added explicit null/omitted legacy-role coverage and proxy switches while cancellation is pending, before optimistic writes.

Validation: 123 focused tests, changed-file TypeScript/lint, and React Doctor100. Review focus: primary versus proxy role semantics at every guard boundary. The API deployment remains unchanged; only the corrected frontend needs another staging deployment before targeted smoke and conditional production promotion.

GelatoGenesis:
The getWalletRole mock is intentional regression coverage: the stale-saved-proxy test sets it to a different profile while the current JWT represents the owner, and verifies that pin/unpin still proceed. Keeping that mock makes the test fail if the original saved-role gate is reintroduced. No runtime dependency on saved role metadata remains.

GelatoGenesis:
Staging validation completed on deployment [34596449445](https://github.com/6529-Collections/6529seize-frontend/actions/runs/34596449445): desktop and 390×844 browser pin/unpin persisted after reload, both test posts were confirmed through the API, original pins were restored, and all 16 missing/invalid-session API cases passed.

The separate [Staging E2E run 34597287670](https://github.com/6529-Collections/6529seize-frontend/actions/runs/34597287670) finished with 186 passed, 34 skipped, and 2 failures (desktop/mobile versions of the same mint-title assertion). `tests/media/media-mint-detail-readonly.spec.ts:121` uses `[^|]+` for the artwork name and rejects the returned title `Mint #547 | The Great Meme Expo | Sgt. Pepe’s World — Episode 1 | The Memes`. This unchanged mint-title test is unrelated to the pin/auth change; recording it separately under the deployment policy. Production API deployment and all 16 production API smoke checks passed before frontend promotion.

PR #3952, opened 2026-09-11
Fix mint E2E title assertion for artwork names containing pipes

GelatoGenesis:
## Issue

The media read-only E2E pack rejects valid mint-page titles when an artwork name contains a pipe. Card #547, “The Great Meme Expo | Sgt. Pepe’s World — Episode 1”, caused the staging and production canaries to fail despite the page rendering correctly.

## Fix

Allow pipes within the artwork-name portion while preserving the anchored mint number and collection suffix, plus the existing plain server-title fallback.

## Changes

- Share the mint-title pattern between the browser assertion and fast Jest coverage.
- Add 12 permanent title-format cases so pipe-containing names remain covered when the live latest artwork changes.
- Frontend test only; no runtime, backend, database, or deployment-order changes.

## Validation

- Reproduced the original assertion failure against production card #547 before editing.
- `6529 run test:e2e:production:media-readonly --reporter=list`: all 5 tests passed after the fix.
- `6529 run test:e2e:staging:media-readonly --reporter=list`: 8 passed across desktop/mobile Chromium; 2 expected skips for the production-only ReMemes fixture.
- `6529 run test:no-coverage __tests__/playwright/mintTitle.test.ts --runInBand`: 12 passed. Restoring the old matcher first reproduced 3 failures for pipe-containing artwork names.
- After sharing the pattern, the targeted production mint browser test passed again; the other media assertions are unchanged.
- `6529 run lint:changed`, `6529 run typecheck:changed`, `6529 run typecheck:playwright`, focused Prettier check, and `git diff --check` passed.

## Risk

- Low: test support, one browser assertion, and focused Jest coverage; application behavior is unchanged.
- Rollback: revert this test-only change if necessary.

## Review Notes

The remaining mint-page content, media, navigation, and metadata assertions are retained. No backend services need deployment. The requested rollout is staging followed by production through the existing frontend workflows.

### DIFF
--- a/__tests__/playwright/mintTitle.test.ts
+++ b/__tests__/playwright/mintTitle.test.ts
@@ -1,0 +1,25 @@
+import { MEMES_MINT_TITLE_PATTERN } from "../../tests/support/mintTitle";
+
+describe("The Memes mint page title", () => {
+  it.each([
+    "Mint | The Memes",
+    "Mint #1 | 6529Seizing | The Memes",
+    "Mint #547 | The Great Meme Expo | Sgt. Pepe’s World — Episode 1 | The Memes",
+    "Mint #12 | One | Two | Three | The Memes",
+    "Mint #12 | Artwork | The Memes | The Memes",
+  ])("accepts the supported title %s", (title) => {
+    expect(title).toMatch(MEMES_MINT_TITLE_PATTERN);
+  });
+
+  it.each([
+    "Mint # | Name | The Memes",
+    "Mint #abc | Name | The Memes",
+    "Mint #1 |  | The Memes",
+    "Mint #1 | Name | Meme Lab",
+    "Other Mint #1 | Name | The Memes",
+    "Mint #1 | Name | The Memes trailing",
+    "Mint | The Memes | extra",
+  ])("rejects the malformed title %s", (title) => {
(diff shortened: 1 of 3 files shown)


GelatoGenesis:
Review follow-up: artwork names are free-form, so a name containing ` | The Memes` is also valid data. The assertion intentionally validates the outer mint-title format; the existing page-content, media, navigation, and mint-metadata assertions remain in place.

The external responsiveness review stopped during setup because it still invokes the removed `6529 install:frozen` command. This test-only diff does not change layout. The corrected staging media pack independently passed in both desktop and mobile Chromium (8 passed, 2 expected production-only skips).

GelatoGenesis:
Added permanent regression coverage in `__tests__/playwright/mintTitle.test.ts`, sharing the same pattern consumed by the browser assertion. The 12 cases cover the plain server title, ordinary and pipe-containing artwork names (including the exact card #547 title), and malformed outer formats. Restoring the original matcher produced 3 failures; the corrected matcher passes all 12 cases. The live production mint browser test also passes after extraction.

Greedy versus lazy matching would not change which anchored titles are accepted here. Artwork names containing the collection phrase remain valid free-form names; the fixed outer suffix and existing page-content checks remain enforced.

PR #3977, opened 2026-09-12
ci: audit independent monitoring provider access

punk6529:
## Issue

Independent operational monitoring needs a verified view of the existing Sentry integration's permissions and monitor availability.

## Fix

Add a manually dispatched, main-only audit using the existing CI token. All Sentry calls are read-only, redirects are rejected, and output is limited to permissions, resource identifiers, response statuses and first-page counts.

## Changes

- No repository checkout or GitHub permissions are needed.
- No token, user details, webhook URLs, vendor response bodies or exceptions are logged.
- The audit does not create monitors or alter provider settings.

## Validation

- `actionlint .github/workflows/monitoring-provider-audit.yml` passed.
- Executed the inline script against read-only fixtures; verified capability detection and that token/vendor payload sentinels do not reach logs or the job summary.
- Whitespace checks passed. Live execution follows merge so the existing secret stays within Actions.

## Risk

- Level: Low. Manual read-only workflow with bounded requests and no application changes.
- Rollback: disable or revert the workflow.

## Review Notes

Check the fixed API destinations and restricted output fields. Counts are explicitly first-page counts; they are not a complete provider inventory.

## Summary by CodeRabbit

* **New Features**
  * Added a manually triggered monitoring access audit for the main branch.
  * Reports authentication status, access scopes, monitor counts, project details, and alert configuration capabilities.
  * Presents sanitized audit results in JSON format within the workflow summary.
  * Clearly indicates missing credentials or unavailable service responses.

### DIFF
--- a/.github/workflows/monitoring-provider-audit.yml
+++ b/.github/workflows/monitoring-provider-audit.yml
@@ -11,10 +11,16 @@ concurrency:
 
 jobs:
   audit:
-    if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
+      - name: Require main
+        shell: bash
+        run: |
+          if [ "$GITHUB_REF" != refs/heads/main ]; then
+            echo '::error::Run this read-only audit from main.'
+            exit 1
+          fi
       - name: Read Sentry monitoring capabilities
         env:
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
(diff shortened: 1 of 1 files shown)


punk6529:
Addressed in the latest commit: the report now exposes specific scope flags and explicitly says these are token metadata, not verified mutation permissions. This avoids conflating the different permissions for project alerts and cron monitors. Raw scope lists are no longer printed.

Added a guarded job-summary path, a sanitized top-level error boundary, and an explicit failure when dispatched outside main. Fixed destinations remain hardcoded; redirects are rejected. Provider bodies are needed to count first-page resources but are never logged. Re-ran actionlint and the read-only/output-redaction fixture successfully.

The responsiveness review is inapplicable to this workflow-only change; no application routes or UI changed.

PR #4006, opened 2026-09-13
Refine marketplace layouts on narrow screens

punk6529:
## Issue

Narrow layouts still show a few presentation defects after #4004: Gradient metadata can overflow when the desktop sidebar remains visible, the NextGen logo and collection control collide at 320px, and prices with different lengths can sit on different rows in Collect's two-column mobile grid.

## Fix

Allow the two artwork headers to reflow and use the existing container-query pattern to give narrow Collect cards a consistent price row followed by their selection action.

## Changes

- Let Gradient metadata cells shrink within their available grid width.
- Let the NextGen brand and collection control wrap when needed.
- Stack card price/actions below a 16rem card width; preserve the existing wider layout.

## Validation

- CollectArtworkCard: nine tests passed; the two header suites: 20 tests passed.
- Changed-file lint and TypeScript checks passed.
- React Doctor: 98/100, no errors; three existing Gradient component warnings.
- Whitespace check passed. Independent source review found no blocker.
- Eleven isolated rendered checks passed: Collect, Gradient and NextGen at 320px, 390px and 1440px, plus Gradient/NextGen with the desktop sidebar at 320px. Document width fits every viewport. Neighboring mobile prices/actions align, narrow headers reflow, and wider layouts remain inline.

## Risk

- Level: Low. CSS class changes in three components; transaction state, exact amounts, selection callbacks and accessible controls are unchanged.
- Rollback: revert this commit through the normal deployment workflow.

## Review Notes

Focus on narrow desktop reflow and consistent mobile price/action alignment. This is the responsive follow-up to the already validated marketplace release in #4004.


punk6529:
Verified both findings against generated CSS and rendered behavior. The installed container-query plugin honors the configured `tw-` prefix: the emitted `.tw-\@container\/artwork` declaration is `container: artwork / inline-size`, and `@container artwork (min-width: 16rem)` sets `flex-direction: row` and `justify-content: space-between` for the two responsive classes.

Browser computed styles confirm `container-name: artwork` and `container-type: inline-size`. Cards around 173px wide in the 390px two-column layout use column direction, with aligned neighboring prices/actions. The 288px cards at 320px single-column width and roughly 349px desktop cards use row direction. All eleven rendered checks pass, including narrow desktop reflow.

Keeping `flex-wrap` is intentional: it remains useful when the row direction activates above 16rem and long labels need additional space. It does not conflict with `flex-direction`; its presence in the smaller column layout is harmless. No syntax or behavior change is needed for these findings.

PR #4010, opened 2026-09-14
Gate version updates and add desktop and mobile update controls

prxt6529:
## Issue

Production version checks could fall back to an individual server's version when release announcements failed. Native pull-to-refresh did not load available app updates, and prominent update toasts covered content without feedback through the reload.

## Fix

Require a valid ready release announcement in production and share update availability between the update control and pull-to-refresh. Native app users get a rocket in a curved attachment above Home. Desktop browsers get an Update row above Search in the lower sidebar; mobile browsers retain the full toast. All three controls show localized loading feedback while reloading the current route.

## Changes

- Mount the rocket inside the navigation dock so its curved attachment moves, hides, and scales with the dock. Home remains a separate navigation action.
- Replace the desktop toast with the existing rocket above Search, using the standard sidebar row's spacing, hover/focus treatment, collapsed Update tooltip and expanded Update label. Mobile browsers, including tablets in desktop mode, retain the full toast.
- Show a static rocket, Updating to the latest version on one line without an ellipsis, and three animated dots on desktop, mobile web and the native app. Respect reduced motion.
- Embed a small transparent WebP of the current green rocket directly in the reload screen HTML, removing the separate image request during document replacement. Prepare the image when an update is available and wait for it before showing the outgoing cover, with a one-second fallback if decoding stalls. A source-hash test keeps the embedded artwork synchronized with the public rocket.
- Restore the loading screen from a short-lived session marker during HTML parsing, before React hydration. Once the incoming app shell is ready, wait for the rocket to decode and paint before clearing the whole cover. Keep a 30-second escape timeout for failed startup. Blocked storage does not prevent reloading.
- Suppress production updates for missing announcement configuration, invalid endpoints, failures, timeouts, and unready responses. Non-production environments without an announcement endpoint retain live-instance comparison.
- Full reload on native pull-to-refresh when an update is available; ordinary content refresh otherwise. Preserve the route and wave reading position, removing only the preview flag.
- React to `showNewVersionToast=true` arriving through client-side deep-link navigation after mounting.
- Update translations, navigation documentation, and the Help Bot corpus.

## Validation

Focused component and reload lifecycle tests cover the desktop row's placement, expansion, availability and localized labels; mobile/native visibility; the dock action, Home navigation, pull-to-refresh, marker consumption and expiry, blocked storage, duplicate taps, route preservation, and timeout recovery. An isolated browser fixture exercises a real document reload with JavaScript deliberately delayed: the cover appears before hydration and clears on shell mount. Phone and tablet geometry checks verify separate Home and rocket targets in expanded and compact states.

The desktop sidebar change uses the existing row component and was checked in focused tests; visual device verification remains manual. The fixture does not reproduce a native WebView's transition before incoming HTML arrives; that gap still needs phone verification. Production build validation requires the checkout's runtime environment configuration.

## Risk

Shared production update availability fails closed when announcements cannot be confirmed. The existing announcement publishing and version-ordering policy is unchanged. The loading cover uses a synchronous static inline bootstrap so it can precede hydration; its timeout exposes the underlying page or error if startup does not complete. Reverting the PR restores the previous update prompt and refresh behavior.

## Summary by CodeRabbit

* **New Features**
  * Added a compact green **Update** pill for mobile and native app users.
  * Selecting the update control reloads the current route while preserving navigation and reading position where possible.
  * Pull-to-refresh now reloads the app when a new version is available; otherwise it refreshes content normally.
  * Added localized update labels in English, French, Spanish, and German.

* **Bug Fixes**
  * Version updates are no longer shown when announcement checks fail or return invalid results.

* **Documentation**
  * Updated mobile navigation, pull-to-refresh, and sharing guidance.


prxt6529:
Follow-up review of `e97657ee382b7c24017f164d513d514f6408d76e`:

- The full reload on a completed pull when an update is available is the requested behavior. Replacing it with cache invalidation would not load the new app bundle. The existing route/wave-position preservation remains in place; this does not promise preservation of all in-memory state or successful loading while offline.
- The forced-toast query is an explicit preview override. The refresh helper removes it with `history.replaceState` before reloading. The earlier sticky-state speculation does not establish a new regression. Missing-announcement logging and blanket docstring coverage are deferred as optional observability/documentation suggestions; production suppression is intentional and covered.
- The [responsiveness run](https://github.com/6529-Collections/6529reviewbot/actions/runs/34813346828) remains a verification gap. It captured blank mobile routes during compilation, unavailable native plugin shims, HTTP 429 responses, and screenshot timeouts. Its logs do not establish that this feature caused the failures, so no unrelated shell, API, or test-runner changes were made. This run is not being represented as passed or as proof that mobile behavior is ready.

The current-head [production build and app checks](https://github.com/6529-Collections/6529seize-frontend/actions/runs/34813032481) completed successfully, providing remote validation where local runtime configuration was unavailable. The branch is zero commits behind `main`; no code change or push was needed for this maintenance pass.

prxt6529:
Reviewed CodeRabbit's two outside-diff comments at head `7686832cc5e64d3b6b67f7ef14e23d315f18233d`. The wave-sharing clipboard behavior in `hooks/waves/useWaveShareCopyAction.ts` and the `waves.drop.open-copy-links` Help Bot entry are unchanged from this PR's current `main` base. This PR changes only the mobile-navigation and pull-to-refresh corpus entries. These wave-sharing findings are deferred to the wave-sharing scope; no unrelated implementation or documentation fixes are included here.

There are no unresolved inline review conversations. The existing optional docstring/observability deferrals remain unchanged.

PR #4017, opened 2026-09-14
fix: increase Jest typecheck heap locally and in CI

prxt6529:
## Issue

The Jest typecheck exhausted the default Node.js heap during a local run.

## Fix

Set `--max-old-space-size=6144` directly on the Node.js process in both `typecheck:jest` and `typecheck:jest:update-baseline` in `package.json`. This applies the same 6144 MiB old-space limit locally and in CI.

## Changes

- Apply the heap limit to both entry points for `scripts/typecheck-test-ratchet.cjs`.
- Keep the existing command guards and baseline-update behavior.
- Use the existing CI `typecheck:tests` command without a workflow-level or step-level heap override. Playwright typechecking remains unchanged.
- Update `pr-reports/4017.md` to describe the final change.

## Validation

- Package JSON and workflow YAML parsed successfully; focused checks confirmed only the two intended package commands changed and the workflow matches the base version.
- `git diff --check` passed.
- The user reran the previously failing local Jest typecheck with `NODE_OPTIONS=--max-old-space-size=6144`; it passed with 1959 diagnostics in 807 files accepted by the existing ratchet.
- Source inspection confirms the ratchet loads TypeScript and calls `ts.createProgram()` and `ts.getPreEmitDiagnostics()` in the same Node.js process. It does not spawn a compiler child process.
- The updated package command itself was not separately executed locally; the successful retry used the equivalent heap setting via the environment.

## Risk

Low: the change raises the old-space ceiling only for the Jest ratchet process. A memory-heavy run can consume more RAM. Remove the flags from the two package scripts to roll back.

## Review Notes

The final implementation is in the package scripts, so CI and local runs share the setting. No staging or production deployment is needed.


prxt6529:
The child-process concern in https://github.com/6529-Collections/6529seize-frontend/pull/4017#issuecomment-5663061828, carried forward in https://github.com/6529-Collections/6529seize-frontend/pull/4017#issuecomment-5663071953, does not apply to this implementation.

The ratchet compiles in-process:
- [Line 8 loads TypeScript directly](https://github.com/6529-Collections/6529seize-frontend/blob/b609dc7f182a6025ba4f0c45491550cac35e03c5/scripts/typecheck-test-ratchet.cjs#L8).
- [Lines 95–106 call ts.createProgram() and ts.getPreEmitDiagnostics()](https://github.com/6529-Collections/6529seize-frontend/blob/b609dc7f182a6025ba4f0c45491550cac35e03c5/scripts/typecheck-test-ratchet.cjs#L95-L106) in that same process.
- The script does not import child_process or spawn tsc. Both package entry points apply the flag directly to the Node.js process performing the compilation.

The user also reran the previously failing local command with NODE_OPTIONS=--max-old-space-size=6144 and obtained: “Jest typecheck ratchet passed: 1959 diagnostics in 807 files.” This verifies that the larger limit was sufficient for that run; the updated package command itself was not separately run locally.

The flag is therefore on the correct process, and no implementation change is required for this finding. Please treat this concern as resolved based on the source evidence. The PR description now reflects the final package-script approach and validation evidence.

PR #4022, opened 2026-09-14
fix: unblock multi-card purchase checks

punk6529:
## Issue  Multi-card Collect purchases can fail before wallet approval because the browser RPC provider rejects requests above 16 KiB. A reproduced four-card request was about 22.5 KiB, producing a generic checks error even though the transaction simulated successfully through the backend provider.  ## Fix  Use an authenticated preflight endpoint for the exact stored batch review. The browser sends only the operation revision and transaction digest, then independently verifies the returned block and applies the existing transaction and fee-cap checks.  ## Changes  - Replace only the batch's oversized browser call and gas-estimation requests. - Preserve complete calldata, recipient, price, wallet and chain validation, send arming, transaction journals and known-hash recovery. - Reject mismatched, stale or malformed preflight responses before wallet approval. - Generate the API models from the canonical schema and update Collect help.  ## Validation  - Nine focused suites: 170 tests passed, including an oversized payload through the real send journal with exactly one mocked wallet request. - Strict changed-file lint and application/test type checks passed; the existing test-type baseline is unchanged. - React Doctor: 100/100, no findings. Final snapshot-boundary follow-up: three affected suites, 69 tests passed. - Exact-head remote CI passed, including the production build, quality/contracts, Collect browser smoke, critical shell, security and installed-app checks. Local build limitations were confined to an unchanged clipboard lint error and dependency-link layout. - Live Rabby/MetaMask transaction completion has not been verified; the regression uses mocked wallet responses.  ## Risk  - Level: Medium. This changes a pre-wallet check in batch purchases. - The API dependency must deploy before this frontend in each environment. Single-item trading and wallet authorization remain unchanged. - Rollback: revert this frontend commit; the additive API remains compatible.  ## Review Notes  Review revision/digest/snapshot binding, exact reviewed caps and preservation of the send journal. No RPC proxy, caller-supplied transaction upload, signing authority or fee-cap increase is introduced.


punk6529:
Reviewed the feedback and added a signed follow-up in d8b90801e06a449897c813df47ab4b0ab304bc46.

- Missing snapshot: intentional fail-closed behavior. The backend batchOperationDto emits transaction, mirror terms and all snapshot fields in the same prepared conditional. An unprepared operation cannot be sent. New regressions explicitly remove each reviewed snapshot field and verify no arm or wallet request occurs; using zero as a fallback would weaken that boundary.
- Digest: both paths call the same marketTransactionDigest helper with the same five transaction fields. The primary execution regression already asserts the preflight digest equals createMarketSendAttempt(...).digest, so that equivalence is tested without creating a second implementation.
- Provider lag: now bounded in both directions at 120 seconds. New tests cover accepted ±12/±120-second differences and rejected ±121-second differences, preserving normal provider lag while rejecting incoherent snapshots.

Follow-up validation: 69 tests across the affected execution/send/cap suites, strict lint, application/test type checks, and React Doctor all passed. The original broader 170-test batch/recovery pack also passed before this focused follow-up.

PR #4025, opened 2026-09-14
Isolate wallet configuration in local Museum publication tests

punk6529:
## Issue

The local anonymous Museum publication pack can fail when Reown's hosted AppKit configuration endpoint returns HTTP 500, even though the Museum page renders. This occurred in the Walker Art Center check in [run 34853823845](https://github.com/6529-Collections/6529seize-frontend/actions/runs/34853823845).

## Fix

The institutional-practice pack now supplies the installed AppKit SDK's valid empty feature configuration for the exact hosted configuration GET, only when testing a local HTTP loopback origin. Hosted wallet features are outside this anonymous publication check.

## Changes

- Add the narrowly scoped fixture and install it before this pack's page navigation.
- Keep staging/production requests, wallet-focused tests, other endpoints/methods, mutation guards, and all console/HTTP error assertions unchanged.
- Add fixture and guard regression tests. Existing Museum profile titles are unchanged.

## Validation

62 focused tests pass, along with Playwright type-checking, changed-file lint, formatting and whitespace checks. An offline Chromium check confirms the local GET is fulfilled, while explicit staging, first-party failures, other wallet endpoints and non-GET requests retain the existing handlers and error assertions.

The full Museum pack has not yet been rerun with this change. The earlier, separate rights-page readiness failure remains unproven; this fix addresses the reproduced Reown request failure.

## Risk

Low: test-only, limited to one local anonymous pack. Remove the fixture and its registration to revert. No application deployment or provider setting change is required.

## Review Notes

Please verify the local-environment boundary and exact endpoint/method match. Live wallet and deployed-environment checks continue to exercise their real requests.


punk6529:
Final-source validation is complete at `afbaf460c986e93f844359d47195442100f6d049`: the full quality/type checks and both Museum desktop/mobile packs passed in run34866687930. The earlier four Playwright mock type errors are fixed; no type baseline or error assertion was weakened.

Disposition of the advisory items:

- Empty `PLAYWRIGHT_ENV` is treated like an unset value. Interception still requires an HTTP loopback base URL and this one anonymous Museum suite. It cannot activate for a hosted staging/production base URL, or an explicit staging/production environment. HTTPS loopback is deliberately excluded from this narrowly scoped fixture.
- The suite clears route registrations before each test and restores the environment after the file completes. Existing calls read only the registration created within that test; no shared asynchronous or concurrent test use is introduced.
- The username/password checks are useful: `URL.origin` omits userinfo, so origin equality alone would not reject a URL containing credentials. Those guards should stay.

CodeRabbit's displayed success context includes a review-rate-limit notice; it is not being counted as a fresh substantive review. The independent source review, 62 focused tests, full type checks and successful desktop/mobile runs provide the validation for this test-only patch. No application deployment is required.

PR #4046, opened 2026-09-15
Audit current Sentry alert routing and credential capabilities

punk6529:
## Issue

The monitoring audit can list legacy Sentry issue rules, but that representation omits current alert conditions and actions. It also cannot distinguish the existing build credential's capabilities when the monitoring credential takes precedence.

## Fix

Read the current workflows for the two monitored projects, their connected detectors, and the existing credentials' capabilities. Emit bounded, sanitized evidence that operators can use to review duplicate notifications and routing before making a separate settings change.

## Changes

- Add bounded workflow/detail/detector reads, configuration fingerprints, recipient kinds, trigger and frequency conditions, and explicit project/connection evidence.
- Resolve supported custom-app targets using the provider's identifier namespaces and classify callback destinations without printing URLs or secrets.
- Check the existing build credential separately with only three capability GETs. Unsupported semantics and incomplete access remain explicitly unverified.
- Preserve manual main-only execution, no checkout or application deployment, fixed API origins, streaming/read budgets, and GET-only provider access.

## Validation

- All 80 tests of the actual inline workflow passed, including privacy canaries, pagination boundaries, provider response shapes, ambiguous target joins, configuration drift, and separate credential handling.
- Windows-aware whitespace check passed.
- Local lint could not run because the existing secure package-install guard rejected differing path/descriptor device IDs for `.npmrc`. The guard and dependency policy were preserved; lint remains for the existing Ubuntu CI checks.

## Risk

- Level: Low. This changes an operator-invoked read-only workflow and its tests. It does not change Sentry settings or application behavior.
- Rollback: revert these two files. No provider-state rollback is needed.

## Review Notes

Check that provider schema assumptions never turn unknown conditions, destinations, or effective permissions into verified claims. Complete inventory reads do not establish event forwarding, notification delivery, or equivalence between legacy rule IDs and workflow IDs.


punk6529:
Reviewed the read-budget findings against the actual output contract. Exhaustion returns `status: "unavailable"`, while a provider rejection retains its numeric HTTP status. Scope and project-access flags become `"unknown"` when the reads are unavailable; they do not become denied permissions. The optional build-credential result does not overwrite the primary credential's evidence or claim effective write access.

The shared 48-read cap and deadline are deliberate overall limits. The per-list and detail limits are additional ceilings, not a guarantee that every maximum-sized combination will fit. A larger configuration can correctly produce an incomplete audit; we will retain that limitation rather than raise the cap or infer missing permissions. The backend project detail is read only after the legacy inventories complete.

Adding an explicit reason alongside `unavailable` could improve diagnostics, but does not change the current decision boundary. I am treating that suggestion and the helper-order/hash micro-optimizations as nonblocking. No provider settings or permissions are changed by this PR.

punk6529:
Checked the additional findings at `4eef1fb76bb7`:

- The fixed project-filtered Sentry listing excludes detached workflows in the current provider implementation. An empty matching inventory remains visible as count zero and empty arrays. Connection-consistency flags do not assert an active route; event forwarding and effective write access remain explicitly unverified. Any later settings decision must inspect the actual connections.
- Build-capability reads run **after** `await configuration(...)`. Identical credentials are explicitly reported as `source: "same_as_selected"`, with no second read. No new secret or warning is needed to make that result visible.
- Destination category and trusted collector equality answer different questions. A recognizable Discord URL is not proof of an approved collector or delivery. Query-bearing URLs cannot match the trusted collector pin, and destination/signing/forwarding verification remains false.
- Rejecting pagination that introduces new fixed query parameters is intentional. Unknown configuration fields participate in the opaque drift fingerprint but are not declared semantically verified. Null fingerprints already prevent list/detail equality and complete audit evidence.
- `setWorkflow` replaces both the list and detail fixture entries. The existing drift test deliberately changes only the detail and asserts failure; the suggested fixture mismatch is not present.

These findings do not identify a blocking error in this GET-only audit. The limited scope and retained unknowns will remain part of the live-result interpretation.

PR #4048, opened 2026-09-15
Match Sentry routing slug checks to audit fingerprints

punk6529:
## Issue

The read-only routing plan stops at `APP_DRIFT` even when its app-slug pin came from the audit. The audit hashes the slug as a canonical JSON string; the routing guard incorrectly compared that pin with a raw-string hash. The stopped plan made no provider changes.

## Fix

Use the audit's canonical fingerprint format for both the installed app slug and webhook donor target comparisons. All pinned values and the remaining drift guards stay unchanged.

## Changes

- Correct two routing comparisons from raw hashing to canonical fingerprinting.
- Align synthetic fixture pins with the audit format.
- Add a cross-workflow regression that executes the actual audit projection, supplies its result to the actual routing plan, and rejects a raw-string pin.

## Validation

- Before the fix: 23 failures among 62 routing scenarios, including the new cross-workflow regression.
- After the fix: all 142 combined audit and routing tests pass, using synthetic data with no external requests.
- Ubuntu CI passed all 142 combined tests, changed-source lint, both typechecks, filesystem/related checks and deployment workflow contracts. SonarCloud and all JavaScript, Python and Actions security/quality scans passed. Whitespace checks passed; broader application build and E2E results report separately.
- The provider-audit workflow and CI wiring are unchanged. No live plan, apply, or provider changes performed for this correction.

## Risk

Low: the change corrects the representation used by two existing comparisons; all 20 pins and the strict preflight/readback behavior remain intact. Reverting this correction restores the earlier fail-closed mismatch. No application deployment is needed.

## Review Notes

Confirm both producers and consumers use SHA-256 over the canonical JSON string, while identifiers and endpoint pins retain their existing hashing conventions. Live configuration still needs verification by the corrected read-only plan before any authorized apply.


punk6529:
The cross-workflow test intentionally depends on the actual audit projection so a producer-contract change cannot silently leave a self-consistent routing fixture passing again. The extraction assertions identify a missing inline block or main boundary, and the known JSON-string vector identifies a changed projection. I’m retaining that explicit dependency without another abstraction or source edit. Ubuntu has now passed all 142 combined tests and changed-source lint.

PR #4055, opened 2026-09-16
Prevent desktop layout flash during native startup

prxt6529:
## Issue

Native launches briefly show the desktop sidebar before React restores the Capacitor layout. Public rendering now starts before wallet initialization, while the hydration-safe device hooks initially report web.

## Fix

Keep public server rendering and wallet-independent startup. Detect iOS/Android synchronously in the document head, then conceal only the initial web layout in native runtimes until the native layout mounts. A dark loading shell reserves header and bottom-navigation space during that interval.

## Changes

- Recognize the Capacitor runtime, custom platform, and native bridges without cookies or user-agent guesses.
- Preserve server content for hydration; hide pending native content from display, accessibility navigation, and keyboard focus.
- Reveal native content based only on layout selection, with no wallet readiness gate or timer.
- Keep standalone access/restricted pages immediately available; they never mount desktop navigation. Cover their hydration and subsequent native navigation on iOS and Android.
- Preserve the version-update reload screen and its handoff to native startup.
- Update the navigation documentation and Help Bot knowledge.

## Validation

- 42 focused Jest tests passed, covering native detection, actual layout hydration, public rendering without wallet startup, and version reload behavior.
- Four additional iOS/Android access/restricted hydration regressions passed, including direct navigation into the native layout, alongside the existing layout and wallet-independent rendering suites.
- 28 browser checks passed across mobile Chromium, iOS WebKit, desktop Chromium, and desktop WebKit. These cover pre-hydration visibility, focus, viewport dimensions, and version-reload handoff using the real boundary/bootstrap/styles. The standalone fixtures include mobile viewport metadata; omitting it reproduced the 980px-versus-390px failure in both native-pending cases.
- Changed-file lint/typecheck and E2E selector checks passed.
- Full-repository lint passed during the production build. Compilation could not complete: Turbopack rejected the worktree's shared node_modules symlink because it points outside the project root. Production compilation remains unverified locally.

## Risk

Medium: this touches initial layout presentation. Physical-device cold starts and the complete app launch have not been exercised locally; browser checks cover the isolated real boundary and Jest covers actual layout selection/hydration. No server or deployment was started. Rollback is removal of the native bootstrap/boundary while retaining the public-rendering architecture.

## Review Notes

The inline script is intentionally synchronous: deferring it through next/script would allow the desktop shell to paint before the Next runtime arrives. The existing hydration-safe device hooks and Wagmi/AppKit startup logic remain unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added a native startup loading shell for iOS and Android.
  * Prevented the desktop sidebar from appearing during native app startup.
  * Preserved native header and bottom-navigation space while the app layout loads.
  * Allowed public content to appear before wallet startup completes.
  * Improved startup transitions to avoid desktop-layout flashes.

* **Bug Fixes**
  * Improved hydration reliability across native access and restricted pages.
  * Preserved version-update reload handoff behavior.

* **Documentation**
  * Added guidance describing native app launch and loading behavior.


prxt6529:
Addressed the access/restricted review concern in 40155e5b152e8533841b1098def0ba4bbaa03dba. The exclusion is intentional: app/access/page.client.tsx and app/restricted/page.client.tsx render standalone login/restriction content, and LayoutWrapper returns those children without ever mounting WebLayout, SmallScreenLayout, or MobileLayout. There is no desktop navigation to conceal on these routes; adding the pending guard would unnecessarily hide their initial content.

Added four iOS/Android regression cases using the real layout selection and device hooks. They verify SSR content survives hydration, every hydration commit has no app-layout ancestor, the startup guard stays absent, and subsequent navigation to /about mounts only the native layout. All 22 tests in the focused layout/public-rendering suites passed; changed-file lint/typecheck and React Doctor passed. The source comment and PR report now explain the exclusion.

The optional detection/interpolation suggestions require no behavior change: only ios/android can activate the guard, and the browser fixture's values are fixed literals. The display: contents wrappers introduce no containing boxes. Physical-device cold starts and full production compilation remain the documented validation follow-ups; this pass changes only regression coverage and explanatory comments/documentation.

PR #4062, opened 2026-09-17
Run full Jest coverage in four parallel CI shards

prxt6529:
## Issue

The full Jest/coverage workflow runs about 20,000 tests in one job. In [this completed main run](https://github.com/6529-Collections/6529seize-frontend/actions/runs/35107527939), Jest took 21m42s of the job's 22m38s.

## Fix

Run the complete suite across four parallel Jest shards, then merge raw Istanbul coverage counters and apply the existing coverage floor. Restore Jest's transformation cache between runs using separate shard keys tied to the toolchain/configuration.

## Changes

- Keep the existing main, coverage-gate PR, and manual triggers, fork policy, coverage provider, source selection, baseline, and tolerance.
- Preserve every shard's outcome with `fail-fast: false`. The existing aggregate check fails unless all four shards succeed, independently of coverage.
- Require all four coverage artifacts; missing, malformed, or empty inputs fail the merge. Merge counters before calculating percentages so complementary hits and untested files are counted correctly.
- Retain separate test/coverage outcome reporting and the combined coverage-summary artifact.

## Validation

- 89 focused tests passed across coverage-floor and testing-strategy; the changed coverage suite was rerun after the final test typing adjustment.
- A small real Jest fixture produced exactly identical full-run and four-shard coverage summaries, including an untested file.
- Full Knip, changed-file lint, and focused Jest TypeScript validation passed. Removed the now-unused coverage-script exports and obsolete Istanbul dependency exception identified by the quality job.
- No browser behavior changes; E2E execution is not applicable. Full-suite timing is intentionally left to this PR's automatically triggered workflow; no local full-suite run or CI polling.

## Risk

Medium: CI scheduling and coverage aggregation change. Four runners can reduce elapsed time but add setup/artifact overhead and may use more runner minutes; actual improvement is not yet measured. Reverting restores the single-job workflow without changing the baseline.

## Review Notes

Check failure propagation and coverage aggregation. Compare this PR's cold-cache duration, and later warm-cache runs, with the existing main run before considering trigger changes.


prxt6529:
Addressed the failed quality job: Knip passes on the PR base, while this PR exposed two unused exports from `coverage-floor.cjs` and made the `istanbul-lib-coverage` dependency exception obsolete. Removed the unused exports and exception; the full Knip check and 26 focused coverage/workflow tests now pass locally.

Review dispositions:
- The shard-result diagnostic already uses `if: always()`, so it runs after a failed merge or download. Keeping it after coverage collection preserves independent test and coverage outcomes. Added a comment explaining this. Bash brace expansion also occurs regardless of whether the files exist; the missing-file error contains the exact input path, now explicitly asserted by the test.
- Both jobs reject fork PRs with the same repository condition. An unexpectedly empty shard deliberately fails rather than silently accepting incomplete execution; added a comment documenting that policy. The existing workflow tests cover fork gating and success/failure/cancelled/skipped result handling.
- Updated the header and missing-summary guidance to describe both single-run and sharded usage.
- Retained Istanbul's standard summary format instead of adding a custom schema/provenance field. The real Jest parity check verified identical summaries from the full-run and merged-shard paths, including untested files.

PR #4070, opened 2026-09-18
Fix Museum Collection aliases with HTTP redirects

prxt6529:
## Issue
Production Museum Rights E2E repeatedly records aborted settings and country-check requests while opening the legacy Collection accession URL. The live alias returns a streamed HTTP 200 shell with a meta refresh; the destination's rights text and layout assertions pass, but errors from the abandoned shell fail the console assertion. See [production run 35330850451](https://github.com/6529-Collections/6529seize-frontend/actions/runs/35330850451).

## Fix
Serve the Collection alias through a GET route handler that returns an empty HTTP 308 before any application shell renders. Preserve the accepted-publication lookup, canonical Work destination, and 404 rejection of unknown or non-Collection objects. Responses are not cached. The Work-to-study browser journey also waits for the link's client-side handler before clicking and verifies that navigation preserves the document.

## Changes
- Replace the redirect-only page with an HTTP route handler; canonical Work rendering and metadata remain on the destination.
- Require the existing desktop/mobile Rights journey to observe the real 308 and final canonical URL. Keep all content, overflow, console, and 5xx checks.
- Cover HTTP response semantics, unavailable publications, rejected aliases, and loader failures. Update the Museum surface registry and page-only route inventory without removing browser coverage.
- Preserve base-registry ownership for paths Git confirms were deleted, so page-to-handler conversions pass change-set validation without ignoring removals or accepting unmapped paths. Regression coverage verifies both registered and unregistered deletions.

- Mark the study link ready after hydration and require the Inside the System journey to preserve `performance.timeOrigin`. This detects a full-document navigation that would cancel startup requests; the server-rendered anchor, destination, copy and styling are preserved.

## Validation
- Passed 90 focused registry, source-shell corpus, local-country fixture, legacy-route and canonical-metadata tests after merging current main.
- Passed changed-file lint/type checking, the complete `typecheck:tests` gate (Jest diagnostic ratchet and Playwright), React Doctor, Museum surface registry, and route inventory checks. The 40 affected tests also pass after correcting fixture types.
- The preceding PR head passed CI's production build and the desktop/mobile Rights journeys on their first attempts. Its later Inside the System journey recorded startup-request cancellations during the study-link click; that separate failure motivated the hydration readiness change.
- Added server-rendering and hydration tests for the study link; all 24 focused link, object-page and registry tests pass, as does the complete Jest/Playwright typecheck gate. The new browser navigation assertion still needs execution in CI. Staging must serve this branch to validate the new 308 contract; resolving the intermittent production cancellations still needs reproduction under comparable conditions or repeated clean first-attempt production Rights runs after deployment.
- No destination layout, copy, artwork, or visual component changes.

## Risk
Medium: changes the transport of one dynamic legacy route family. Invalid aliases return a plain HTTP 404; valid aliases retain the same destination and membership policy. Rollback is a revert. The cancellation mechanism is strongly supported by production evidence, but a staging pass alone does not establish that production flakiness is resolved.

## Review Notes
Current main, including #4051, is merged into this branch. Its country-check fixture remains local-only; staging and production use live requests. No settings/country failures are allowlisted, and no retries or pack selection are relaxed.

## Summary by CodeRabbit

* **New Features**
  * Legacy Museum collection URLs now permanently redirect to canonical Work pages before rendering.
  * Added reliable Museum study links that preserve destinations and styling during hydration.

* **Bug Fixes**
  * Unknown, unavailable, or non-Collection object URLs now return uncached 404 responses.
  * Improved accession and system deep-link navigation, including correct redirect destinations and browser behavior.

* **Tests**
  * Expanded coverage for redirects, canonical metadata, deep links, hydrated navigation, and route handling.


prxt6529:
Follow-up disposition of the advisory review:

- HEAD is preserved: Next.js 16.2.11 automatically implements HEAD using GET when a route handler exports GET but no HEAD (`next/dist/server/route-modules/app-route/helpers/auto-implement-methods.js`). The earlier HEAD→405 observation is incorrect. Unsupported mutation methods remain rejected.
- The plain 404 for unknown/non-Collection aliases is an intentional, documented tradeoff of the transport-only handler. The previous Museum not-found component uses `DEFAULT_LOCALE`, not request-selected localization. This change does remove the styled not-found shell and its navigation link for invalid legacy aliases; keeping that limitation explicit in the PR risk notes is appropriate. Restoring a rendered shell or adding a second redirect/error route would expand this fix. The string Response uses the standard `text/plain;charset=UTF-8` default.
- The optional hostile-objectId test expansion is deferred: input is used for the existing accepted-publication lookup, never interpolated into Location. `museumWorkHref` validates canonical Work IDs and encodes them under `/museum/network/works/`; the fallback also requires that canonical prefix. Existing contract coverage rejects unknown and non-Collection aliases and asserts the exact allowed redirect destination.

The strengthened 308 assertion proves the transport change. A staging pass alone does not prove the intermittent production cancellations are resolved; that still requires reproduction under comparable conditions or repeated clean first-attempt production Rights results after deployment.

PR #4077, opened 2026-09-20
Fix Collect CI fixture for artwork documentation lane

punk6529:
## Issue

The Collect CI contract test omits `playwright_artwork_documentation` from its mock plan. The workflow now reads that lane's `required` flag, so the unrelated-documentation case throws instead of reaching its assertions.

## Fix

Add the missing lane with `required: false`, matching an unrelated documentation change.

## Changes

One test-fixture entry. Existing assertions, workflow behavior, and application code are unchanged.

## Validation

- Focused `collect-browser-ci.test.ts`: 8 tests passed.
- `lint:changed`, `typecheck:changed`, and whitespace checks passed.
- Reviewed the workflow's browser-lane selection and relevant pack registration. No Playwright or app build was run because this correction only completes the Jest mock; browser behavior and selection rules are unchanged.

## Risk

Low: test-only fixture correction. Reverting it would restore the failing test setup.

## Review Notes

The unrelated-documentation case must continue to assert that neither Collect nor core browser work is required.

## Summary by CodeRabbit

- **Tests**
  - Updated the plan fixture to include Playwright artwork documentation as an optional check.


punk6529:
@coderabbitai review

punk6529:
Review follow-up: the real effective planner emits `checks.playwright_artwork_documentation` in `scripts/app-pr-ci-effective-plan.cjs`, and the workflow publishes that check and reads it when selecting the smoke lane. The fixture therefore needs the key even when it is false.

Positive and negative lane selection is already covered in `__tests__/scripts/app-pr-ci-effective-plan.test.ts`, including artwork, layout and browser-policy paths plus unrelated documentation. The existing Collect assertions stay unchanged. The focused Collect suite passes all eight tests, and PR quality/contracts, CodeQL, Sonar, Snyk and DCO passed on this head.

The generic docstring warning concerns an existing test helper whose mock object gains one property; adding a docstring or restructuring that fixture is outside this correction. No application behavior, workflow, dependency or deployed asset changes in this PR.

